PR #10, opened 2013-03-29
Enable flag on exposures to optionally exlcude nil attributes.

idyll:
For certain APIs I find that it is best for the client if the API hide's attributes that are nil.

Instead of handling this in each serializer (json, plist, xml) I am looking for options to pull control over this into the entity.


yeonhoyoon:
As for now, you can do something like this.
```ruby
with_options({ 
  if: lambda { |object, options|
    current_attr = options[:attr_path].last
    !object[current_attr].nil?
  }
}) do
  expose :a
  expose :b
end
```

PR #45, opened 2014-01-08
Ability to "flatten" nested entities into parent (e.g. for CSV)

joelvh:
This feature allows you to specify another entity to copy exposures from. It's useful when flattening a hierarchy for CSV. It also adds the `:object` option to exposures that can define an alternate object to get values from. (Specifying an alternate object is necessary when copying exposures because the exposures don't refer to the original entity/model anymore.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ This gem adds Entity support to API frameworks, such as [Grape](https://github.c
 ```ruby
 module API
   module Entities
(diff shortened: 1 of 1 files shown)


brianphillips:
:+1: on getting the `merge_with` bits pulled in

dblock:
@brianphillips Want to try and extract a `merge_with` PR out of this ?

PR #99, opened 2014-11-21
raising ArgumentError if exposure_options[:using] does not respond to represent

themusicman:
The current NoMethodError that gets raised requires time to trace and debug. The ArgumentError that is raised now gives a little more information and context to fix the issue quicker.


dblock:
Bump.

PR #110, opened 2015-03-10
Fixed exposure of attributes marked as safe when using hash models.

croeck:
The valid_exposure check now also accepts Hash objects that have a key matching the attribute's name.
Fix includes `spec` that demonstrates the issue and fails without the changes.


dblock:
Thanks.

PR #111, opened 2015-03-10
Fix: allow usage of attributes with name 'key' if `Hash` objects are used

croeck:
Fix now is checking the method arity in the `delegate_attribute` method. With this change the key attribute will be fetched and instead of the `key` method (with arity 1) being invoked.
The fix includes a `spec` that demonstrates the issue and fails without the changes.

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -556,17 +556,17 @@ module Grape
       name = self.class.name_for(attribute)
       if respond_to?(name, true)
         send(name)
+      elsif object.is_a?(Hash)
+        object[name]
+      elsif object.respond_to?(name, true)
+        object.send(name)
+      elsif object.respond_to?(:fetch, true)
+        object.fetch(name)
       else
-        if object.respond_to?(name, true)
+        begin
           object.send(name)
-        elsif object.respond_to?(:fetch, true)
-          object.fetch(name)
-        else
-          begin
-            object.send(name)
-          rescue NoMethodError
-            raise NoMethodError, "#{self.class.name} missing attribute `#{name}' on #{object}"
-          end
+        rescue NoMethodError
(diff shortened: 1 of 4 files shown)


lib/grape_entity/entity.rb
```diff
@@ -557,7 +557,7 @@ def delegate_attribute(attribute)
       if respond_to?(name, true)
         send(name)
       else
```
dblock: While you're here, I think this can be collapsed into an `elsif`.

lib/grape_entity/entity.rb
```diff
@@ -557,7 +557,7 @@ def delegate_attribute(attribute)
...
-        if object.respond_to?(name, true)
+        if object.respond_to?(name, true) && object.method(name).arity == 0
```
dblock: This worries me. If there's a special case with `Hash` I would just check for that.

croeck:
Alright, I'll update the pull request in a few minutes...

croeck:
I would fail fast if the `object` is as `Hash` and does not contain the `name` as key.
Would this be ok?

    def delegate_attribute(attribute)
      name = self.class.name_for(attribute)
      if respond_to?(name, true)
        send(name)
      elsif object.is_a?(Hash)
        raise NoMethodError, "#{self.class.name} missing attribute `#{name}' on #{object}" unless object.key? name 
        object[name]
      elsif object.respond_to?(name, true)
        object.send(name)
      elsif object.respond_to?(:fetch, true)
        object.fetch(name)
      else
        begin
          object.send(name)
        rescue NoMethodError
          raise NoMethodError, "#{self.class.name} missing attribute `#{name}' on #{object}"
        end
      end
    end

croeck:
Pull is updated, arity check was replaced with `object.is_a? Hash` and `if..else` was combined to one large `if..elsif..else` statement

dblock:
Much better. Merging.

PR #114, opened 2015-03-13
Added 'only' option that selects which attributes should be returned

estevaoam:
This PR intends to add an option named `only`, when representing an object or collection.

The motivation for this is that we use both `grape` and `grape-entity` intending to creating a well-structured and performant API. 

Some great service APIs, like [Facebook](https://developers.facebook.com/docs) and [Github](https://developer.github.com/v3/) ones allow the consumer for the API to choose which fields he wants in this request. Also, this [article](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#limiting-fields) is a very interesting read.

Based on this, I developed this feature into `grape-entity`, and I would love to check some opinion about this.

Thanks. :smiley:


README.md
```diff
@@ -220,6 +220,41 @@ class MailingAddress < UserData
...
+  expose :email
+end
+
+class Entity
+  expose :id
+  expose :title
+  expose :user, using: UserEntity
+end
+
+data = Entity.represent(model, only: 'name,user.name,user.email')
```
dblock: I really think that `only` should take an array or a single symbol, not a comma-delimited string.

dblock:
I really like this, except for the `only` DSL. I think it should be strongly typed, example:

```
Entity.represent(model, only: [:name, user: [:name, :email]])
```

It's easy to turn a comma-delimited string above into something like this, and if you accept such a string you should be really validating its syntax, which I feel is outside of the scope of this gem.

You also don't need to support a comma-delimited version at all, Rack supports passing arrays in, so you could make a query like `only[]=name&only[user][]=name&only[user][]=email`. It feels a bit ugly but it's standard (and you can make it less ugly by having a middleware that does the parsing into params.

estevaoam:
@dblock thanks for your comment. I agree with you, I was blind in the mindset of an API paying attention about how I needed to pass to grape-entity so I can have this feature.

You're right, it makes more sense that some middleware takes care of it instead of the gem.

I'm gonna work on those changes. Thanks for your insight!

estevaoam:
@dblock When you have a chance, could you take a look on my modifications? Thanks!

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ============
 
 * Your contribution here.
+* [#114](https://github.com/intridea/grape-entity/pull/114): Added 'only' option that selects which attributes should be returned
```
dblock: Swap this with the line above.

README.md
```diff
@@ -220,6 +220,41 @@ class MailingAddress < UserData
...
+{
+  title: 'grape-entity is awesome!',
+  user: {
+    name: 'John Applet',
+    email: 'john@example.com'
+  }
+}
+```
+
+Instead of returning all the exposed attributes. You can use this to build a great API service
```
dblock: I think you should remove the "You can use this ... to the performance.", it's kinda a judgement call.

dblock:
This looks good. Please fix the build and squash the commits. To fix the build just do a `rubocop --auto-gen-config`.

estevaoam:
@dblock done. :+1:

dblock:
Merged via https://github.com/intridea/grape-entity/commit/bf0084729ca1ea73a488ccf0cd19e453b3472ca7. 

For next time, Rubocop's workflow is like this:

```
rubocop -a # auto-correct
rubocop --auto-gen-config # update config with any unfixed violations
```

PR #116, opened 2015-03-19
Root node inheritance

guidoprincess:
Currently, exposure is inherited from parent entities to child entities.  So, something like this works:

```
module Foo
  module Entities
    class GenericEntity < Grape::Entity
      expose :id
    end
  end
end

module Foo
  module Entities
    class SpecificEntity < GenericEntity
      #has :id already exposed
    end
  end
end
```

However, "root" is not inherited, so the following does not work

```
module Foo
  module Entities
    class GenericEntity < Grape::Entity
      root :data, :object
    end
  end
end

module Foo
  module Entities
    class SpecificEntity < GenericEntity
      #has no knowledge of parent's root settings
    end
  end
end
```

This pull request addresses the above, which is useful for those of us who want to have one standard root node name throughout the application and don't want to define it in every entity.

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -407,17 +407,27 @@ module Grape
     # @option options :only [Array] all the fields that should be returned
     def self.represent(objects, options = {})
       if objects.respond_to?(:to_ary) && ! @present_collection
-        root_element =  @collection_root
+        root_element =  root_element(:collection_root)
         inner = objects.to_ary.map { |object| new(object, { collection: true }.merge(options)).presented }
       else
         objects = { @collection_name => objects } if @present_collection
-        root_element =  @root
+        root_element = root_element(:root)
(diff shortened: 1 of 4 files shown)


lib/grape_entity/entity.rb
```diff
@@ -420,6 +420,16 @@ def self.represent(objects, options = {})
       root_element ? { root_element => inner } : inner
     end
 
+    # This method returns the entity's root or collection root node, or its parent's
+    # @param root_type: either :collection_root or just :root
+    def self.root_or_super(root_type)
```
dblock: Nitpick: I find the method name unnecessarily convoluted. Why don't we just make this `root_element` or `root`?

dblock:
This is good. See my comment on method naming and please squash the commits. Thanks!

guidoprincess:
Done!

dblock:
Merged.

PR #117, opened 2015-03-20
Fix: using 'only' option with exposed attribute that is an entity

estevaoam:
I found a bug when using the recent added `only` option, when trying to return attributes that was evaluated as grape entity, this is the fix for it.

I’m not sure about if should be added as a fix in the Changelog because the version isn’t released yet, if I need to change something just let me know.

Thanks!


lib/grape_entity/entity.rb
```diff
@@ -650,7 +650,14 @@ def options_for_using(attribute, options)
...
       using_options[:root] = nil
-      using_options[:only] = only_fields(using_options)[attribute]
+      attribute_only_option = only_fields(using_options)[attribute]
+
+      if attribute_only_option.is_a?(Array)
+        using_options[:only] = attribute_only_option
+      else
+        using_options.delete(:only)
+      end
+
```
dblock: You should sink the checking for `Array` inside `only_fields`, this way you don't need to check anything here. If that function returns `nil` you'll be assigning `using_options[:only] = nil` which is what we want.

dblock:
No need to add to CHANGELOG, indeed.

estevaoam:
@dblock did the changes you suggested, when you can, please take a look. :+1:

lib/grape_entity/entity.rb
```diff
@@ -499,8 +499,8 @@ def should_return_attribute?(attribute, options)
       only_fields(options).include?(self.class.key_for(attribute))
     end
 
-    def only_fields(options)
-      return {} unless options[:only]
+    def only_fields(options, for_attribute = nil)
+      return nil unless options[:only]
```
dblock: Pedantic change, you don't need to `return nil`, just `return`, `nil` is by default.

dblock:
Good, merging.

lib/grape_entity/entity.rb
estevaoam: Thanks for the tip, sorry, not used to use only return. :)

Issue #121, opened 2015-04-10
Subclasses keep the parent's .documentation?


uppfinnarn:
I'm not sure if I'm using it wrong, or if it's bugged, but if I do this:

```ruby
# api/entities/my_entity.rb
module API
  module Entities
    class MyEntity < Grape::Entity
      expose :name, documentation: { type: String, required: true }
      expose :message, documentation: { type: String }
      
      expose :created_at, documentation: { type: DateTime, required: true }
      expose :updated_at, documentation: { type: DateTime, required: true }
      
      expose :relation, using: API::Entities::Other, documentation: { is_array: true }
      
      def self.entity_name() "MyEntity" end
      
      class Update < API::Entities::MyEntity
        unexpose :created_at
        unexpose :updated_at
        unexpose :relation
      end
    end
  end
end

# api/entities/other.rb
module API
  module Entities
    class Other < Grape::Entity
      expose :created_at, documentation: { type: DateTime, required: true }
      expose :updated_at, documentation: { type: DateTime, required: true }
    end
  end
end
```

The results aren't quite what you'd expect in regards to `.documentation`:

```
$ rails console
Loading development environment (Rails 4.2.1)
irb(main):001:0> API::Entities::MyEntity.documentation
=> {:name=>{:type=>String, :required=>true}, :message=>{:type=>String}, :created_at=>{:type=>DateTime, :required=>true}, :updated_at=>{:type=>DateTime, :required=>true}, :relation=>{:is_array=>true}}
irb(main):002:0> API::Entities::MyEntity::Update.documentation
=> {:name=>{:type=>String, :required=>true}, :message=>{:type=>String}, :created_at=>{:type=>DateTime, :required=>true}, :updated_at=>{:type=>DateTime, :required=>true}, :relation=>{:is_array=>true}}
```

As a result, it currently appears to be impossible to use Entities for parameter reuse, without resorting to copypasting `expose …` lines into a separate class - making `unexpose` rather useless.

As a side note, it doesn't appear to be `unexpose` itself that's broken, as `expose` in a subclass similarly has no effect.

dan-corneanu:
Can you please describe what result are you expected from the above setup?

dan-corneanu:
Can you please describe what result are you expected from the above setup?

uppfinnarn:
I would expect `unexpose`'d attributes in the subclass to be absent from the subclass' `documentation`, and vice versa for newly `expose`'d.

[...]

PR #122, opened 2015-04-11
Subclass documentation #121

dan-corneanu:
This is a proposed fix for issue
Subclasses keep the parent's .documentation? #121

Currently each Entity keeps track of its own exposures in a class object instance variable. A new entity class makes a copy of it's parent exposures then goes on and modifies its own exposures (add or remove exposures).

For the moment, I think we can fix the documentation issue by simply removing the call to **superclass.documentation**.

In the future we might have to replace the class object instance variable (which is a simple hash at the moment) with something more elaborate (like a chained filter or something) to handle the child - parent relationship better. Ex. what happens if we add some more exposures to the parent entity after we defined the child entity? I think that the child entity will not catch that extra exposure. 

Please review this pull request and let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 * [#114](https://github.com/intridea/grape-entity/pull/114): Added 'only' option that selects which attributes should be returned - [@estevaoam](https://github.com/estevaoam).
 * [#115](https://github.com/intridea/grape-entity/pull/115): Allowing 'root' to be inherited from parent to child entities - [@guidoprincess](https://github.com/guidoprincess).
+* [#121](https://github.com/intridea/grape-entity/pull/122): Sublcassed Entity#documentation properly handles unexposed params - [@dan-corneanu](https://github.com/dan-corneanu).
 * Your contribution here.
 
 0.4.5 (2015-03-10)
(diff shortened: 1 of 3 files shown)


dblock:
This looks good. Please update CHANGELOG and squash the commits.

dan-corneanu:
Thank you for reminding me.
Done

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 * [#114](https://github.com/intridea/grape-entity/pull/114): Added 'only' option that selects which attributes should be returned - [@estevaoam](https://github.com/estevaoam).
 * [#115](https://github.com/intridea/grape-entity/pull/115): Allowing 'root' to be inherited from parent to child entities - [@guidoprincess](https://github.com/guidoprincess).
+* [#121](https://github.com/intridea/grape-entity/pull/122): Sublcassed Entity#documentation properly handles unexposed params. - [@dan-corneanu](https://github.com/dan-corneanu)
```
dblock: The period goes at the end, not after params :)

dblock:
Can you please squash these and correct the super minor thing in CHANGELOG? Thanks.

dan-corneanu:
squashed

dblock:
Merging, thanks.

Issue #56, opened 2014-02-13
Fix or workaround for duplicate exposure caveat?


nickls:
I'm looking to expose using different entities if a user is anonymous or not:

```
class MediaEntity < Grape::Entity
  expose :id
  expose :user, using: Entities::UserEntity, if: lambda { |object, options| object.anonymous == false }
  expose :user, using: Entities::AnonymousUserEntity, if: lambda { |object, options| object.anonymous == true }
end

present :data, media, with: MediaEntity
```

This doesn't work as you documented in the caveat section. Is there a good workaround?

Also, do you have a solution in mind for this?

Morred:
@nickls @donbobka @dblock Is this still being worked on, or can I help in any way? I ran into exactly the same problem today and would love to see this working (unless there's a specific reason for the exposures overwriting each other).

dblock:
Would love a pull request that addresses the spec failure mentioned above.

Morred:
Cool, I'll see what I can do.

Morred:
I took @donbobka's test and adapted it a bit: https://github.com/Morred/grape-entity/tree/double-exposure

The test fails, and what happens is technically exactly what you're describing in the caveat section of the Readme, the second exposure overwrites the first exposure. Don't know if this is considered a bug (because you do document it in the Readme), but it would certainly be a lot more convenient if this did not happen.

Morred:
Completely forgot the pull request >.< It's there now.

marshall-lee:
I don't see a bug here. Both https://gist.github.com/nickls/9560205 and #124 demonstrate documented behavior: exposure with the same name overwrite previously defined.

However, It would be good to not overwrite them. I think `Grape::Entity`'s DSL for exposures is very similar to a regular ruby code so principle of least surprise says us that DSL is expected to behave as a regular Ruby code. For example:
```ruby
# Grape::Entity:
expose :a
expose :a, if: -> (object, options) { object.some_condition? }

# Regular ruby code:
expose object.a
expose object.a if object.some_condition?
```

Looks similar, right?

I definitely want this feature to be implemented and I think I could do it. But it will break backward compatibility (i.e. that *caveat* in README about double exposures will gone) and it requires big changes in code.

So what should be done in examples:

```ruby
expose :x, using: E1
expose :x, using: E2
```
Here `x` should be exposed using `E2`.

```ruby
expose :x, using: E1, if: -> { some_condition? }
expose :x, using: E2
```
[...]

PR #124, opened 2015-04-29
Add failing test for double exposures #56

Morred:
#56

### DIFF
--- a/spec/grape_entity/entity_spec.rb
+++ b/spec/grape_entity/entity_spec.rb
@@ -276,6 +276,58 @@ describe Grape::Entity do
           expect(subject.represent(object).send(:value_for, :size)).to eq object.class.to_s
         end
       end
+
+      context 'duplicate exposure names with conditional' do
+        it 'should return the correct result' do
+          class PhotoEntity < Grape::Entity
+            expose :photo_file
+          end
+
+          class SoundEntity < Grape::Entity
+            expose :sound_file
+          end
+
+          class MediaEntity < Grape::Entity
+            expose :media, using: SoundEntity, if: lambda { |object, _| object.sound? }
+            expose :media, using: PhotoEntity, if: lambda { |object, _| object.photo? }
+          end
+
+          class Photo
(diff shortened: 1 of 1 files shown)


marshall-lee:
Hello, @Morred!

I closed & reopened your PR just to initiate a re-run of Travis CI checks. Build still fails but this time it's just due to RuboCop offenses. RSpec tests pass fine as you can see here: https://travis-ci.org/intridea/grape-entity/jobs/73909288 so issue #56 can be considered as solved.

I will not merge this PR because new specs from #151 for this issue are sufficient. If you don't think so and want this to be merged please somehow fix RuboCop offenses.
```
spec/grape_entity/entity_spec.rb:338:52: C: Use the new lambda literal syntax ->(params) {...}.

            expose :media, using: SoundEntity, if: lambda { |object, _| object.sound? }

                                                   ^^^^^^

spec/grape_entity/entity_spec.rb:339:52: C: Use the new lambda literal syntax ->(params) {...}.

            expose :media, using: PhotoEntity, if: lambda { |object, _| object.photo? }

                                                   ^^^^^^
```

I personally don't like this type of RuboCop offense about lambda syntax and if you think the same feel free to open a separate PR with patch for `.rubocop.yml`.

I also don't like defining constants like `PhotoEntity` in specs because these things are global and survive among rspec examples. I know there are some specs in `entity_spec.rb` that already define constants but it's something that I want to get rid of. I already had issues with constants while working on this project. @dblock, what do you think? Also I need to say that this problem is gracefully solved in https://github.com/Casecommons/with_model gem but it's only for ActiveRecord models. I also seen a good workaround for this in `Her` gem, please look at https://github.com/remiprev/her/blob/49cd15ffbad2aec104810753a90d9a7b6977fd31/spec/support/macros/model_macros.rb#L6 and https://github.com/remiprev/her/blob/c39fbb6d211109da5f2ff8c35fff4b5a85d8d5c9/spec/spec_helper.rb#L21

PR #134, opened 2015-06-04
Some subtle improvements

etehtsea:
Because less code is better than more.

### DIFF
--- a/spec/grape_entity/entity_spec.rb
+++ b/spec/grape_entity/entity_spec.rb
@@ -296,21 +296,10 @@ describe Grape::Entity do
           expect(subject.exposures).to eq(name: {}, email: {})
         end
 
-        # the following 2 behaviors are testing because it is not most intuitive and could be confusing
-        context 'when  called from the parent class' do
-          it 'remove from parent and all child classes that have not locked down their attributes with an .exposures call' do
+        context 'when called from the parent class' do
+          it 'remove from parent and do not remove from child classes' do
             subject.expose :name, :email
             child_class = Class.new(subject)
-            subject.unexpose :email
-
-            expect(subject.exposures).to eq(name: {})
-            expect(child_class.exposures).to eq(name: {})
-          end
-
-          it 'remove from parent and do not remove from child classes that have locked down their attributes with an .exposures call' do
-            subject.expose :name, :email
-            child_class = Class.new(subject)
-            child_class.exposures
             subject.unexpose :email
 
             expect(subject.exposures).to eq(name: {})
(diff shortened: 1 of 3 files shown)


spec/grape_entity/entity_spec.rb
```diff
@@ -296,24 +296,13 @@ class Parent < Person
...
+          it 'remove from parent and do not remove from child classes' do
             subject.expose :name, :email
             child_class = Class.new(subject)
             subject.unexpose :email
 
             expect(subject.exposures).to eq(name: {})
-            expect(child_class.exposures).to eq(name: {})
-          end
-
-          it 'remove from parent and do not remove from child classes that have locked down their attributes with an .exposures call' do
```
dblock: Can you explain why the other test is no longer necessary?

etehtsea: Because there is no more "confusing" behaviour. `unexpose` no longer affects child classes regardless of was `.exposures` method called in them or not.

dblock: I am confused, does this spec break now? I mean when you unexpose :email from the parent class, the child no longer has it?

In which case this at least needs an UPGRADING entry, although I think you want to track down the PR in which this was added and figure out why?

etehtsea: from now, if you unexpose from parent class, child doesn't affect.

**master:**
```
[3] pry(main)> parent = Class.new(Grape::Entity)
=> #<Class:0x0000010250cf00>
[4] pry(main)> parent.expose :name, :email
=> [:name, :email]
[5] pry(main)> child = Class.new(parent)
=> #<Class:0x00000102a25938>
[6] pry(main)> child.exposures
=> {:name=>{}, :email=>{}}
[7] pry(main)> child_leaking = Class.new(parent)
=> #<Class:0x0000010292fab0>
[8] pry(main)> # not called .exposures
[9] pry(main)> parent.unexpose :email
=> {}
[10] pry(main)> child.exposures
=> {:name=>{}, :email=>{}}
[11] pry(main)> child_leaking.exposures
=> {:name=>{}}
```

**my-branch:**
```
[1] pry(main)> parent = Class.new(Grape::Entity)
NameError: uninitialized constant Grape
from (pry):1:in `__pry__'
[2] pry(main)> require 'grape-entity'
=> true
[3] pry(main)> parent = Class.new(Grape::Entity)
=> #<Class:0x000001031fb108>
[4] pry(main)> parent.expose :name, :email
=> [:name, :email]
[5] pry(main)> child = Class.new(parent)
=> #<Class:0x000001032a0888>
[6] pry(main)> child.exposures
=> {:name=>{}, :email=>{}}
[7] pry(main)> child_leaking = Class.new(parent)
=> #<Class:0x0000010319f588>
[8] pry(main)> # not called .exposures
[9] pry(main)> parent.unexpose :email
=> {}
[10] pry(main)> child.exposures
=> {:name=>{}, :email=>{}}
[11] pry(main)> child_leaking.exposures
=> {:name=>{}, :email=>{}}
```

It was introduces in f089f16e and I think comment 

```
# the following 2 behaviors are testing because it is not most intuitive and could be confusing
```

explains why. Because exposures merge happened in children classes only when it was called.

I'll add entry to `UPGRADING`.

etehtsea: I didn't find UPGRADING, so added note to CHANGELOG

dblock:
Ok, I am good with this, merging. Thank you very much.

dblock:
Check out https://github.com/intridea/grape-entity/pulls too, there're a few bugs that were raised that sound vaguely familiar, maybe this fixes anything.

marshall-lee:
I think more and more that #161 and #166 are caused by this patch.

#166 is easy to fix
but for #161: `superclass` traversal was an intended thing but now it's removed. Issue with `unexpose` on parent class should be fixed different way.

PR #136, opened 2015-06-24
Allow for strings in 'only' and 'except' options

bswinnerton:
With the addition of the `only` and `except` options, possibilities such as the [partial response](http://bit.ly/1fxWuXJ) design pattern (where the consumer of the API can specifically request only the fields that it desires) become available. A common pattern to request these fields would be via `params`, which will arrive as strings. Currently only an array of symbols (or hash containing symbols) is supported to define the desired fields.

As discussed [here](https://github.com/intridea/grape-entity/pull/114#issuecomment-79122967), it certainly should be the responsibility of another gem to do parameter validation in a partial response scenario, but such a gem would also have to do a [perhaps unnecessary] recursive symbol coercion as well when it could be handled by grape-entity, as proposed by this pull request.

This commit updates Grape::Entity to support an array of symbols _or_ strings to be passed as the `only` or `except` options of `represent`.

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -465,7 +465,7 @@ module Grape
         else
           allowed_fields[attribute] = true
         end
-      end
+      end.symbolize_keys
 
       if for_attribute && @only_fields[for_attribute].is_a?(Array)
         @only_fields[for_attribute]
@@ -486,7 +486,7 @@ module Grape
         else
           allowed_fields[attribute] = true
         end
-      end
+      end.symbolize_keys
 
       if for_attribute && @except_fields[for_attribute].is_a?(Array)
         @except_fields[for_attribute]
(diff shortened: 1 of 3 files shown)


lib/grape_entity/entity.rb
```diff
@@ -454,16 +454,17 @@ def should_return_attribute?(attribute, options)
     end
 
     def only_fields(options, for_attribute = nil)
-      return nil unless options[:only]
+      return unless options[:only]
 
       @only_fields ||= options[:only].each_with_object({}) do |attribute, allowed_fields|
         if attribute.is_a?(Hash)
           attribute.each do |attr, nested_attrs|
+            attr = attr.to_sym
```
dblock: I think you should convert those to strings. Until Ruby 2.2 symbols are not garbage collected so you're creating an opportunity for some DDoS if this is used with user-supplied data.

dblock:
See my comment around symbols vs. strings to be converted/used internally. Otherwise this makes total sense.

bswinnerton:
@dblock what are your thoughts on [this](https://github.com/bswinnerton/grape-entity/commit/7814c1d855596e7398cd746c80b9c84e52ed9e08) approach (and trailing method call on an `end` style)? 

Rather than coercing the keys to strings inside the loop, we can use ActiveSupport's `symbolize_keys` after the appropriate hash of keys has been built. I _think_ this should solve the garbage collection issue you brought up.

dblock:
I think this is fine. 

I'd like to have explicit tests for this rather than updating existing tests with mixed keys. Write one that says something like "it works for strings", another that "it works for symbols" and another that "it works for mixed strings and symbols".

bswinnerton:
Absolutely. Just added explicit tests for the symbol and string code paths of `only` and `except`, each with nested attributes.

dblock:
Good. Could you please squash these commits? I'll merge.

bswinnerton:
Much thanks @dblock! Done.

dblock:
Merged.

PR #139, opened 2015-07-09
tracking attribute path into options[:attr_path]

calfzhou:
Make Grape::Entity keep tracking of current being exposed attribute's full path into `options[:attr_path]`.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -7,12 +7,12 @@
 
 # Offense count: 8
 Metrics/AbcSize:
-  Max: 51
+  Max: 53
 
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 328
+  Max: 346
 
 # Offense count: 5
 Metrics/CyclomaticComplexity:
@@ -26,7 +26,7 @@ Metrics/LineLength:
 # Offense count: 7
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 32
+  Max: 37
 
 # Offense count: 5
 Metrics/PerceivedComplexity:
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -428,22 +428,11 @@ module Grape
       opts = options.merge(runtime_options || {})
 
       root_exposures.each_with_object({}) do |(attribute, exposure_options), output|
-        next unless should_return_attribute?(attribute, opts) && conditions_met?(exposure_options, opts)
-
-        partial_output = value_for(attribute, opts)
-
-        output[self.class.key_for(attribute)] =
-          if partial_output.respond_to?(:serializable_hash)
-            partial_output.serializable_hash(runtime_options)
-          elsif partial_output.is_a?(Array) && partial_output.all? { |o| o.respond_to?(:serializable_hash) }
-            partial_output.map(&:serializable_hash)
-          elsif partial_output.is_a?(Hash)
-            partial_output.each do |key, value|
-              partial_output[key] = value.serializable_hash if value.respond_to?(:serializable_hash)
-            end
-          else
-            partial_output
-          end
+        parent_path = track_attr_path(attribute, opts)
+        if should_return_attribute?(attribute, opts) && conditions_met?(exposure_options, opts)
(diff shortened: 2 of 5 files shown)


dblock:
I really like this, needs a CHANGELOG entry, please.

lib/grape_entity/entity.rb
```diff
@@ -425,22 +425,27 @@ def serializable_hash(runtime_options = {})
...
-          elsif partial_output.is_a?(Array) && !partial_output.map { |o| o.respond_to?(:serializable_hash) }.include?(false)
-            partial_output.map(&:serializable_hash)
-          elsif partial_output.is_a?(Hash)
-            partial_output.each do |key, value|
-              partial_output[key] = value.serializable_hash if value.respond_to?(:serializable_hash)
+        parent_path = track_attr_path(attribute, opts)
+        begin
+          next unless should_return_attribute?(attribute, opts) && conditions_met?(exposure_options, opts)
+
+          partial_output = value_for(attribute, opts)
```
dblock: Maybe construction of `partial_output` can be refactored into a method?

calfzhou:
squashed and rebased to current master.

dblock:
Merged.

.rubocop_todo.yml
```diff
@@ -7,12 +7,12 @@
 
 # Offense count: 8
 Metrics/AbcSize:
-  Max: 51
+  Max: 53
```
marshall-lee: Metrics should not increase. Why we use RuboCop if we just stab it?

spec/grape_entity/entity_spec.rb
```diff
@@ -991,6 +996,110 @@ def embedded
...
+            end
+          end
+
+          fresh_class.class_eval do
+            expose :characteristics, using: EntitySpec::CharacterEntity
+
+            protected
+
+            def self.path_for(attribute)
+              attribute == :characteristics ? :character : super
```
marshall-lee: I don't like this idea. This thing should be implemented in exposure parameters. For example:

```ruby
expose :characteristics, using: EntitySpec::CharacterEntity, attr_path: proc { :character }
```
What if we have more attributes which path we need to override? We'll end up with conditional hell in `path_for`.

marshall-lee:
@calfzhou, @dblock 

I backported this feature while working on rebasing #151. But I changed your specs because in the new design there's no place for methods like `key_for`, `name_for` and (newly introduced) `path_for`.

Instead of this I implemented a `:attr_path` exposure option that adds the same possibility:
```ruby
expose :charactersistics, attr_path: -> (obj, opts) { :character }
expose :need_to_skip, attr_path: -> (obj, opts) { nil }
```

Maybe it would be useful to also add these possibilities in feature:
```ruby
expose :need_to_skip, attr_path: :character # just constant
expose :need_to_skip, attr_path: nil # just skip
```

.rubocop_todo.yml
dblock: Personally I don't care as much, I think it's a good flag and if it makes sense, I would not make it a barrier.

spec/grape_entity/entity_spec.rb
dblock: :+1:  on what @marshall-lee says, would love a refactor @calfzhou

dblock:
Thanks @marshall-lee your changes make sense to me.

calfzhou:
@marshall-lee 's `attr_path` exposure is pretty good!

.rubocop_todo.yml
marshall-lee: Yes I think the same but I just want to see things *before* changing `.rubocop_todo.yml`. There could be something we can suggest to easily fix and if there is not then we suggest to just regenerate `.rubocop_todo.yml` :)

marshall-lee: As features and their correctness are superior to style pedantry, I want to push #160.

I think the reason that people just stab RuboCop offenses is that it runs *before* RSpec. It prevents maintainers to just see the feature implemented correctly or not.

dblock: I would like to disable this offense in Rubocop, and be able to run `--auto-gen-config` for everything else. However when you try to do the latter it ignores what's in `.rubocop.yml`, which defeats the whole purpose.

Feel free to switch it to run after, btw.

PR #141, opened 2015-07-16
Some array handling micro optimizations

marshall-lee:
1. `any?` without block will still try to iterate through array looking for non-falsey values. But we just need to check do we have `nested_exposures` or not.
2. `nested_exposures_for?` is introduced. It does a good think because current implementation pessimistically assumes that nested exposures exist.
3. `!map { ... }.include?(false)` is self-explaining. There's a convenient `all?` method.
4. `reject { ... }.select { ... }` can be done with one `select` or with one `reject`. I choosed `select`.
5. `nested_value_for` is extracted because of Rubocop detected offenses. Maybe it's worth to extract other condition branches to separate methods or even to separate classes (replace conditional with polymorphism refactoring)

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -399,8 +399,8 @@ module Grape
     end
 
     def valid_exposures
-      exposures.reject { |_, options| options[:nested] }.select do |attribute, exposure_options|
-        valid_exposure?(attribute, exposure_options)
+      exposures.select do |attribute, exposure_options|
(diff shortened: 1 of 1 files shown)


marshall-lee:
These things improved `bench/serializing.rb` result only by 4 percents. I think it's possible to write a more illustrative benchmark but it's not necessary.

Anyway this patch doesn't change code too much and doesn't reduce readability too much so I think it's acceptable.

dblock:
I like it, merging. Thanks.

Issue #140, opened 2015-07-16
:safe option


marshall-lee:
Hello!

Here's the excerpt from README:

> Don't raise an exception and expose as nil, even if the :x cannot be evaluated.
> expose :ip, safe: true

However, such fields are not exposed as `nil` — they're simply not exposed. Moreover this is written in specs:

```ruby
сontext 'with safe option' do
  it 'does not throw an exception when an attribute is not found on the object' do
    fresh_class.expose :name, :nonexistent_attribute, safe: true
    expect { fresh_class.new(model).serializable_hash }.not_to raise_error
  end

  it "does not expose attributes that don't exist on the object" do
    fresh_class.expose :email, :nonexistent_attribute, :name, safe: true

    res = fresh_class.new(model).serializable_hash
    expect(res).to have_key :email
    expect(res).not_to have_key :nonexistent_attribute
    expect(res).to have_key :name
  end
  # ...
end
```

Who's right — the README or the specs?
My opinion that this is a bug. If we say `expose` then it should be somehow exposed. Not exposing it is just surprising.

@dblock 
I want to fix it because this `:safe` thing stops me from fixing another bugs and it stops me even from reasoning about that bugs :smile: But what to fix — implementation+broken specs or just README ?

dblock:
Obviously README and implementation should match. I think attribute that doesn't exist vs. `:x` cannot be evaluated are different things though, and I would definitely welcome a README update on what happens when the attribute doesn't exist vs. cannot be evaluated.

marshall-lee:
@dblock 
> attribute that doesn't exist vs. :x cannot be evaluated are different things

What is "cannot be evaluated" then? When it raises exception? It seems like this option wasn't intented for such things. I think it's implemented for exposing mixed collections of objects — some of them may have some attribute and some of them may not.

If I want expose something *optionally* I would use `:if` or `:unless`. If exposure is not conditional then it should be exposed anyway and if it cannot (`respond_to?` if false or `NoMethodError` is raised) then raise an error (if unsafe) or expose `nil` (if safe).

For example:

```ruby
class E < Grape::Entity
  expose :a, if: -> (obj,opts) { obj.respond_to? :a }
  expose :b, safe: true
end

# ...

X = Struct.new(:a, :b)
Y = Struct.new(:a)
Z = Struct.new(:b)

# ...

E.represent(X.new(1,2), serializable: true) # should evaluate to { a: 1, b: 2 }
[...]

PR #142, opened 2015-07-16
Fix safe private method calls

marshall-lee:
Related to #140

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -589,7 +589,7 @@ module Grape
       (nested_exposures.any? && nested_exposures.all? { |a, o| valid_exposure?(a, o) }) || \
         exposure_options.key?(:proc) || \
         !exposure_options[:safe] || \
-        object.respond_to?(self.class.name_for(attribute)) || \
+        object.respond_to?(self.class.name_for(attribute), true) || \
         object.is_a?(Hash) && object.key?(self.class.name_for(attribute))
     end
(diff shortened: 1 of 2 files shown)


marshall-lee:
It should be merged only if #140 is a false alarm. Otherwise it should be fixed differently

dblock:
So the problem was private methods? I feel like an error that the method is private is almost better than exposing private methods automatically. What do you think?

Either way this needs a CHANGELOG entry, please.

marshall-lee:
@dblock 

It seems like exposing private methods was intented.
Look at the current [`delegate_attribute` implementation](https://github.com/intridea/grape-entity/blob/35ada1ebbd0659cd7686b7b21796a51b6bf1a6e2/lib/grape_entity/entity.rb#L572)

It checks that object responds to attribute name using `object.respond_to?(name, true)` (not just `object.respond_to?(name)` and evaluates it by `object.send(name)` (not by `object.public_send(name)`).

So this PR just makes safe and unsafe exposures behave identical because now only public methods can exposed when using `:safe` option.

As for me I don't like exposing private methods. If you think that only public methods should be exposed I could open a separate PR for this too. But it probably can be considered as a breaking change. Is it acceptable here?

dblock:
Ok I am with you on this one @marshall-lee. Lets call it a bug and add a line to CHANGELOG please (via `--amend`) that says  something like "Fix: expose works with private methods".

PR #144, opened 2015-07-18
Refactored examples namespacing.

castillobg:
Refactored namespacing to be more consistent with the one used in the [Example](https://github.com/intridea/grape-entity#example): ``API::Entities::Status``, ``API::Entities::Address``, instead of ``API::Status`` or ``API::Address``, to make it explicit that the presenters should also be GrapeEntities.


dblock:
Makes sense.

PR #147, opened 2015-07-20
Refactoring: extract delegating logic.

### LINKED ISSUE #140: :safe option

marshall-lee:
This one solves many problems at once:

 - Fixes #140. For this one I changed specs a bit. It's breaking but I think it's worth it.
 - Fixes #142 differently because now `respond_to?(name, true)` stuff is incapsulated in `Delegator::PlainObject`.
 - Old `delegate_attribute` catched `NoMethodError` and re-raised it with custom message. It's incorrect because `NoMethodError` can occur deep inside in method call — this exception simply doesn't mean that attribute doesn't exist.
 - Solves the problem that object is checked to `is_a?(Hash)` and `respond_to?(:fetch, true)` multiple times.
 - Extracts delegating logic to separate delegator classes.
 - removes constructing of `valid_exposures` at all — there's no need in this method now.


dblock:
I like this, merging as is.

PR #151, opened 2015-07-21
Refactoring: extract exposures.

### LINKED ISSUE #56: Fix or workaround for duplicate exposure caveat?

### LINKED ISSUE #112: Nested Exposure attributes are flatten in documents
the [Nested Exposure](https://github.com/intridea/grape-entity#nested-exposure) generates incorrect documentation, all nested fields are flatten
```ruby
expose :name
expose :contact_info do
  expose :phone
  expose :email
end
```

it should generate (document)
```yaml
name: ...
contact_info:
  phone: ...
  email: ...
```

but now it generates
```yaml
name: ...
phone: ...
email: ...
```

### LINKED ISSUE #149: @only_fields and @except_fields memoization is wrong
`only_fields` and `except_fields` methods internally memoizing `@only_fields` and `@except_fields` instance variables respectively. It's wrong because they're calculated against passed `options` parameter which can differ between separate `serializable_hash` calls.

I'd open a PR later.

### LINKED ISSUE #150: :unless condition with Hash: wrong implementation and specs
`:if` it can be used with hash like `if: { cond1: 'foo', cond2: false }`. It means that this exposure should be exposed iff **all the conditions** are matching. It means that it should not be exposed with options like:
 - `cond1: 'bar', cond2: false`
 - `cond1: 'foo', cond2: true`
 -  `cond1: 'foo'`
 -  `cond2: false` .

The only valid `options` value when it's exposable is `cond1: 'foo', cond: false`.

Documentation says that `:unless` is an opposite of `:if`. So the exposure should be exposable iff **any of the conditions** is **not** matching.

And i also found a spec that tells the same idea in its text: *only passes through hash :unless exposure if any attributes do not match*:

```ruby
it 'only passes through hash :unless exposure if any attributes do not match' do
  exposure_options = { unless: { condition1: true, condition2: true } }

  # right
  expect(subject.send(:conditions_met?, exposure_options, {})).to be true
  # wrong
  expect(subject.send(:conditions_met?, exposure_options, condition1: true)).to be false
  # right
  expect(subject.send(:conditions_met?, exposure_options, condition1: true, condition2: true)).to be false
  # wrong
  expect(subject.send(:conditions_met?, exposure_options, condition1: false, condition2: true)).to be false
  # right
  expect(subject.send(:conditions_met?, exposure_options, condition1: true, condition2: true, other: true)).to be false
  # right
  expect(subject.send(:conditions_met?, exposure_options, condition1: false, condition2: false)).to be true
end
```

If you look carefully you'll see It's wrong:

 - expectation with options `condition1: true` should be true, because it doesn't contain matching `condition2` element.
 - expectation with options `condition1: false, condition2: true` should be true because `condition1` doesn't match.

### LINKED ISSUE #152: unexposure is unaware of nesting
```ruby
def self.unexpose(attribute)
  exposures.delete(attribute)
end
```
It just deletes attribute from `exposures` table. Of course, it's possible to call it like `unexpose(:some__nested)` but it's a hack that lacks internal representation which now is going to be removed(#151). It also doesn't clear the `nested_exposures` table which is potentially buggy.

I propose to temporarily forbid calling `unexpose` inside of nesting exposure (with exception) because it's unclear what to unexpose in this case. Current documentation says nothing about the calling `unexpose` withing nesting exposure so I think it's safe.

### LINKED ISSUE #153: @documentation memoization is slightly incorrect
```ruby
def self.documentation
  @documentation ||= exposures.each_with_object({}) do |(attribute, exposure_options), memo|
    if exposure_options[:documentation].present?
      memo[key_for(attribute)] = exposure_options[:documentation]
    end
  end
end
```

It works incorrect in this corner case:
```ruby
entity_class = Class.new(Grape::Entity)
entity_class.expose :x, documentation: { desc: 'just x' }
doc1 = entity_class.documentation
entity_class.expose :y, documentation: { desc: 'just y' }
doc2 = entity_class.documentation
doc1 # => {:x=>{:desc=>"just x"}} 
doc1 == doc2 # => true
```
It could be fixed simply by assigning `nil` to `@documentation` in `expose` and `unexpose`.
Will be fixed in #151

### LINKED ISSUE #155: Serializing of deeply nested presenter exposures
```ruby
it 'serializes deeply nested presenter exposures' do
  e = Class.new(Grape::Entity) do
    expose :y
  end
  subject.expose :a do
    subject.expose :b do
      subject.expose :c do
        subject.expose :lol, using: e
      end
    end
  end
  expect(subject.represent(lol: { y: 123 }).serializable_hash).to eq(
    a: { b: { c: { lol: { y: 123 } } } }
  )
end
```

Now fails with:

```
       -:a => {:b=>{:c=>{:lol=>{:y=>123}}}},
       +:a => {:b=>{:c=>{:lol=>#<#<Class:0x00000003a499c8>:0x00000003a3f0b8 @delegator=#<Grape::Entity::Delegator::HashObject:0x00000003a3f090 @object={:y=>123}>, @object={:y=>123}, @options={:root=>nil, :only=>nil, :except=>nil}>}}},
```

Will be fixed in #151.

### LINKED ISSUE #156: Deep projections (:only, :except) are unaware of nesting
Failing example:
```ruby
context 'with nested attributes' do
  before do
    subject.expose :additional do
      subject.expose :something
    end
  end

  it 'preserves nesting' do
    expect(subject.represent({ something: 123 }, only: [ { additional: [ :something ] } ], serializable: true)).to eq(
      additional: {
        something: 123
      }
    )
  end
end
```

Deep projections work well when passing to presenter exposures (`:using` options) but they should work with nested exposures too.

Will be fixed in #151

marshall-lee:
This one reaches many goals at once:

 - Fixes #56.
 - `exposures` hash table is removed and substituted with
   `root_exposures` array.
 - Due to previous point, tree structure `nested_exposures` based on
   flat hash table with keys like `root_node__node__node` is removed
   because now double exposures don't rewrite previously defined
   so such tree structure is simply incorrect.
   `NestingExposure` with an array of children is introduced instead.
 - Ones who want an old rewriting behavior should manually `unexpose`.
 - Fixes #112.
 - Fixes #149: runtime `options` are now wrapped in an `Options` object.
 - Fixes #150: see new `spec/grape_entity/exposure_spec.rb`.
 - Fixes #152.
 - Fixes #153.
 - Fixes #155.
 - Fixes #156.
 - All exposure configuration and exposing strategies are extracted to
   the separate classes.
 - `key_for`, `name_for` internal methods are removed.
 - Much of the overhead is gone so performance is increased by 20-30%.


marshall-lee:
At first glance this change may seem too big. But every move here is reasonable.
For instance It's impossible to solve #56 without removing `nested_exposures`. Removing `nested_exposures` requires different way to represent nesting trees so I cannot find something better than extract exposure logic from `Entity` at all.

marshall-lee:
But some problem is introduced with changing `exposures` from hash table to an array so `grape-swagger` will not work with this patch. And it's even unclear how to generate documentation. Imagine this case:

```ruby
expose :x, using: E1, if: :condition1
expose :x, using: E2, if: :condition2
```

Which one should the swagger pick: `E1` or `E2` ?

I can help with implementing a workaround for swagger but for now I don't know exactly how  to fix it.

Maybe bring back construction of `exposures` as a hash table and deprecate it?

dblock:
I'd like to merge this, it needs a detailed CHANGELOG entry and we have to discuss what to do about swagger. The conflicting cases should just have some predictable behavior.

marshall-lee:
I just read through swagger specification and didn't find anything that might help gracefully solve this issue.

But:

1. It tries to analyze only those exposures that has a `:documentation` option.
2. It fetches `exposure_options` by name (which now is not unique) to get its `:using` option only if there is no `:type` field in `:documentation` of exposure.
3. `Grape::Entity.documentation` (which is used by `grape-swagger`) still has an old *rewriting* nature. But I think that in this case it's good. Ones who use conditional exposures with different `using` values are forced to specify one `:type` for exposures with the same key (`:as` value or attribute name). This `:type` value could be simply `'object'` or some `'AbstractEntity'`.
4. According to previous point for any key`grape-swagger` should pick `:type` from documentation or `:using` (if there's no `:type`) from the last exposure with this key and it's already implemented in `documentation`.
6. `Grape::Entity.documentation` still has an issue described in #112. I propose to solve it traversing only `root_exposures`, not `exposures`.
7. Remove flat array of `exposures`. There's no need in this now.

I'd like to propose a workaround to `grape-swagger` based on a fact that new entities do not respond to `exposures` method. It would also require adding different versions of `grape-entity` to `grape-swagger`'s travis matrix.

Any complains?

dblock:
No complaints. Grape-swagger needs a lot of work too and as long as we offer people a path forward I'm cool with it. Please at the very least open issues with grape-swagger for anything introduced here.

dblock:
@marshall-lee Lets get this merged? Amend CHANGELOG? We also probably need an UPGRADING document.

marshall-lee:
@dblock I'm working on CHANGELOG right now.
I'd like to write UPGRADING.md too. To what version we're upgrading? 0.5.0? Maybe change `lib/grape_entity/version.rb` in this patch too?

And what format should I use in UPGRADING?

dblock:
0.5.0 makes sense

marshall-lee:
Oops, JRuby build failed. https://travis-ci.org/intridea/grape-entity/builds/72464413

Latest changes revealed some misbehavior in JRuby, I reported it here: https://github.com/jruby/jruby/issues/3178

marshall-lee:
LOL, Now Rubocop suggests me in https://travis-ci.org/intridea/grape-entity/jobs/72467983:
```
Pass &:nesting? as an argument to take_while instead of a block.
```
But `take_while` with `&:method` is buggy in JRuby as I mentioned above.

marshall-lee:
But it's okay on `jruby-head`, lol.

marshall-lee:
@dblock I'm almost ready. Maybe you release 0.4.6 without these changes? Or just put all pending changes to 0.5.0?

dblock:
@marshall-lee You've done some great work here. I emailed you about becoming a maintainer on grape-entity. Your first order of duty could be to release 0.4.6?

u2:
:+1:

rngtng:
great stuff, to bad it breaks https://github.com/Gild/ruby-swagger too. How does the UPGRADE look like? next time pls deprecate the method first..

rngtng:
I think the line @calfzhou is referring to is: https://github.com/ruby-grape/grape-swagger/blob/03ee40d41d71063236c4b0b1d468f69978ab308e/lib/grape-swagger/doc_methods.rb#L261

rngtng:
ok this did the trick for me, super hacky, dirty and for sure not complete... 

```ruby
class BaseDecorator < Grape::Entity
    def self.exposures
      extract_exposures(root_exposures, {})
    end

    private

    def self.extract_exposures(root, hash, prefix = nil)
      root.each do |entity|
        key = "#{prefix}#{entity.attribute}"
        nested = prefix.nil? ? {} : { nested: true }
        if entity.is_a?(Grape::Entity::Exposure::NestingExposure)
          hash[key.to_sym] = {}
          extract_exposures(entity.nested_exposures, hash, key + '__')
        else
          hash[key.to_sym] = entity.send(:options).merge(nested)
        end
      end
      hash
    end
  end
```

dblock:
@rngtng Want to give fixing https://github.com/ruby-grape/grape-swagger/issues/314 a try?

rngtng:
@dblock I'd love to, if I find the time...let me see..

PR #160, opened 2015-08-03
Run rspec before rubocop.

marshall-lee:
I think it's reasonable to run rspec before rubocop.


marshall-lee:
It also does a good job in that it doesn't run RuboCop when specs fail.

PR #165, opened 2015-08-03
Fix: entity instance methods exposure.

marshall-lee:
Port of #164 to master.


spec/grape_entity/entity_spec.rb
```diff
@@ -1467,6 +1467,10 @@ def name
         rep = EntitySpec::DelegatingEntity.new(friend)
         expect(rep.value_for(:name)).to eq 'cooler name'
         expect(rep.value_for(:email)).to eq 'joe@example.com'
+
+        another_friend = double('Friend', email: 'joe@example.com')
+        rep = EntitySpec::DelegatingEntity.new(another_friend)
+        expect(rep.value_for(:name)).to eq 'cooler name'
```
dblock: Generally deserves its own spec.

marshall-lee: I agree. Sorry

PR #170, opened 2015-08-25
Clarify ActiveSupport dependencies

marshall-lee:
Following https://github.com/ruby-grape/grape/pull/1122.


dblock:
Would be curious to find out how this change affects a simple benchmark, too.

marshall-lee:
`$ ruby bench/serializing.rb`:

Before:
```
       user     system      total        real
serializing  1.310000   0.000000   1.310000 (  1.313772)
```

After:
```
       user     system      total        real
serializing  1.260000   0.000000   1.260000 (  1.257941)
```

Became better but not much.
See also https://github.com/ruby-grape/grape/pull/1122#issuecomment-134556477

marshall-lee:
IMO it became better only because I removed `delegate`. It helped me before already.

PR #172, opened 2015-09-15
support dynamic attributes & merging collection values

danhodge:
This is the start of my proposed solution to the problem I outlined on the mailing list [here](https://groups.google.com/forum/#!topic/ruby-grape/cWnj8tZnOMw). It doesn't add general purpose dynamic attribute, it just allows you to evaluate the 'parent' attribute of a nested attribute structure dynamically through a hacky naming convention (`__` prefix on the attribute name). I would have rather used a block to evaluate the value of the dynamic attribute, but I wasn't sure what all of the the implications of creating a `NestingExposure` were.

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -498,7 +498,7 @@ module Grape
 
     # All supported options.
     OPTIONS = [
-      :rewrite, :as, :if, :unless, :using, :with, :proc, :documentation, :format_with, :safe, :attr_path, :if_extras, :unless_extras
+      :rewrite, :as, :if, :unless, :using, :with, :proc, :documentation, :format_with, :safe, :attr_path, :if_extras, :unless_extras, :combine
     ].to_set.freeze
 
     # Merges the given options with current block options.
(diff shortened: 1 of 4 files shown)


dblock:
I am personally not a huge fan of naming conventions. There should be an explicit way to tackle this. Try to propose a cleaner DSL? Maybe something as simple as an `expose` that doesn't take a name? What if the following exposed each field in the hash?

```ruby
expose do
 { x: 1, y: 2 }
end
```

marshall-lee:
Hello!

This feature seems good, but:

>  a hacky naming convention (__ prefix on the attribute name).

Please don't!

It'll be better to implement it in terms of `:as` option:

```ruby
expose :by_date, as: -> (obj, opts) { obj.date } do
  expose :first_name
  expose :last_name
end
```

Also the question is how it will work with `:only` and `:except` options?

marshall-lee:
And: 

> merging collection values

What is this? I don't see any merging in your code.

danhodge:
I only used the naming convention for prototyping purposes, I like the `as: <lambda>` approach much better. I'll switch to that approach. As for the merge, it happens [here](https://github.com/danhodge/grape-entity/blob/6331474abf72244042b4022368331e3ce5ce739a/lib/grape_entity/exposure/base.rb#L68). Instead of mapping over the objects in the collection, it reduces them into a single `Hash` by merging them together. Perhaps `reduce` is a better term for what its doing?

dblock:
@marshall-lee Does the `as: ` syntax let you merge onto root at set of dynamic keys, doesn't look like that or am I missing something?

Can you comment on what you think about the key-less extension (`expose` with just a block)?

PR #173, opened 2015-09-15
Fix class name in docs

iangreenleaf:
A small typo, but an important one. It definitely confused me for a moment.


dblock:
Thanks!

Issue #131, opened 2015-05-19
Problems after update


reiz:
One of my projects is running stable with grape-entity `0.3.0`. But as soon I update to `0.4.0` or higher some of my tests fail. The odd thing is that they fail randomly and it seems that the connection to MongoDB randomly fails. I'm getting errors like this: 

```

Problem:
   Document(s) not found for class Project with id(s) 555b4d484fe53d4ad2000310.
 Summary:
   When calling Project.find with an id or array of ids, each parameter must match a document in the database or this error will be raised. The search was for the id(s): 555b4d484fe53d4ad2000310 ... (1 total) and the following ids were not found: 555b4d484fe53d4ad2000310.
 Resolution:
   Search for an id that is in the database or set the Mongoid.raise_not_found_error configuration option to false, which will cause a nil to be returned instead of raising this error when searching for a single id, or only the matched documents when searching for multiples.
```

If I downgrade to `0.3.0` all tests are green again. 

Any idea what could be the reason?

dblock:
That is very strange indeed. If you can put the project up I can help you debug.

idyll:
I kinda remember having an issue like this...

I think that is do to a change with behaviours around exposing models that may not have the method being exposed. 

Previously this would work as long as it wasn't called, now it throws an exception.

reiz:
Thanks for the input. I will give it a try again on the weekend.

epaulet:
I'm also having issues updating from 0.4.5 to 0.4.8. Before if a model did not have a method, it would that was exposed in an entity, it wouldn't raise any exceptions. Now there is a NoMethodError.

epaulet:
Seems to be that passing in `safe: true` doesn't seem to work in conjunction with `format_with`

marshall-lee:
@epaulet please open a PR with a failing test or post an example here at least.

epaulet:
@marshall-lee I added a PR with a failing spec for this issue I was referring to. Not sure if it's related to the topic of this thread or not.

PR #174, opened 2015-09-17
Be able to use format_with safely

epaulet:
I don't have any proposed solution for this, but this is in response to https://github.com/ruby-grape/grape-entity/issues/131

Currently, `safe: true` does not do anything when used together with `format_with`. Before 0.4.6, it would just return a nil value if a method did not exist. As of 0.4.6, there is a NoMethodError instead.

### DIFF
--- a/spec/grape_entity/entity_spec.rb
+++ b/spec/grape_entity/entity_spec.rb
@@ -1063,6 +1063,20 @@ describe Grape::Entity do
           expect(res).to have_key :name
         end
 
+        context "when used with format_with" do
+          it "does expose attributes that don't exist on the object as nil" do
+            fresh_class.format_with :date_formatter do |date|
+              date.iso8601
+            end
+
+            model = {}
+
+            fresh_class.expose(:timestamp, format_with: :date_formatter, safe: true)
+            res = fresh_class.new(model).serializable_hash
+            expect(res).to have_key :timestamp
+          end
+        end
+
         it 'does expose attributes marked as safe if model is a hash object' do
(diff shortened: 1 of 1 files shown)


marshall-lee:
Thanks, will look at this soon.

marshall-lee:
This spec fails on `0.4.5` too but for different reason.

marshall-lee:
```
     Failure/Error: expect(res).to have_key :timestamp
       expected #has_key?(:timestamp) to return true, got false
```

marshall-lee:
So `0.4.5` doesn't return `nil` value as you mentioned. It simply doesn't expose your attribute at all.

marshall-lee:
...but i agree it's some kind of bug that affects only `Hash` objects. I will fix it soon on both `0-4` and `master`.

epaulet:
@marshall-lee I added some more to this to be more clear about the issue I'm having.

This is the type of behavior I was seeing in previous versions (0.4.5) where a nil value wasn't passed into a formatters when `safe: true` was passed in as an option. This issue still was not resolved with a7d6df4e51dd4ef9515a31d935810e8bd29e18a0

marshall-lee:
@epaulet thanks, will review soon.

epaulet:
The `safe` flag seems to be for exposing `nil` if the entity doesn't have the correct attribute. The intention here is to ignore the attribute altogether if it resolves to nil. Perhaps an `allow_nil` flag would be more appropriate. Either way, going to close this PR for now since the solution seems like the wrong one...

PR #178, opened 2015-09-30
Fixes typo in Entity Organization docs

cakejelly:
Just noticed this earlier when I was searching the docs for something. All of the sample code in this example references a `Status` whereas this line was referencing a `User`, which doesn't make sense. I assume this was a typo? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,7 +414,7 @@ class Status
 end
 ```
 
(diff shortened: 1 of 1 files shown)


dblock:
Yep, thanks!

PR #182, opened 2015-10-13
fix(options): Preserve Only and Except Options

### LINKED ISSUE #181: Only Option Not Correct for Nested Value

Linell:
Don't override `options[:only]` and `options[:except]` when merging in new options.

Resolves #181

### DIFF
--- a/lib/grape_entity/options.rb
+++ b/lib/grape_entity/options.rb
@@ -28,6 +28,13 @@ module Grape
                    else
                      @opts_hash.merge(new_opts)
                    end
+
+          only = opts_hash[:only]
+          merged[:only] = only unless only.nil? || only.empty?
+
+          except = opts_hash[:except]
+          merged[:except] = except unless only.nil? || only.empty?
+
           Options.new(merged)
         end
       end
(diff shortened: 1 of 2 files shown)


dblock:
This seems correct, but we need a test that fails on HEAD currently, please, and a CHANGELOG entry.

lib/grape_entity/options.rb
```diff
@@ -28,6 +28,15 @@ def merge(new_opts)
                    else
                      @opts_hash.merge(new_opts)
                    end
+
+          unless opts_hash[:only].nil? || opts_hash[:only].empty?
```
dblock: I think you want `opts_hash.key?(:only) || ...`

marshall-lee:
Thanks for your contribution!

@dblock please hold it for a while, i want to dig it deeper.

Linell:
@dblock I made the change that you pointed out.

As far as the test goes, I honestly wouldn't know how to write it for this.

dblock:
I'll leave this to @marshall-lee.

@Linell without this fix something breaks in your code, right? Write a test that reproduces the broken behavior without this change.

PR #187, opened 2015-10-27
Allow to associate meta data with entities

dim:
It would be really handy to be able to associate custom meta data with entities to generate automated documentation.


dblock:
I am not against this. You should update README and such and lets see what others say about this feature.

dim:
Thanks, updated the README as suggested

PR #202, opened 2016-01-27
Fix: reset `@using_class` memoization on `.setup`

rngtng:
Fixes a tiny bug where calling `.setup` on `RepresentExposure` wouldn't 'reset' `@using_class`. As of this, it was possible that `@using_class` and `@using_class_name` were referencing different classes.


dblock:
Thanks. I'll deal with the build on rbx separately.

PR #203, opened 2016-01-27
Fix: return `self` instead of `nil` if no attribute was deleted

rngtng:
If a given attribute of `delete_if` couldn't be found, `nil` was returned [as described in ruby spec](http://ruby-doc.org/core-2.2.4/Array.html#method-i-reject-21). Although, I expected `@exposures`  to be return, to maintain consistency with the cases when sth. is found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 * Your contribution here.
 * [#202](https://github.com/ruby-grape/grape-entity/pull/202): Fix: Reset `@using_class` memoization on `.setup` - [@rngtng](https://github.com/rngtng).
+* [#203](https://github.com/ruby-grape/grape-entity/pull/203): Grape::Entity::Exposure::NestingExposure::NestedExposures.delete_if always returns exposures. - [@rngtng](https://github.com/rngtng).
 
 0.5.0 (2015-12-07)
 ==================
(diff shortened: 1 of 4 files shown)


dblock:
Makes sense. This changes the interface, I think we need an UPGRADING ala https://github.com/ruby-grape/grape/blob/master/UPGRADING.md, could you make one please? Bump version to 0.6.0 as well.

rngtng:
done, what u think @dblock ?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 * Your contribution here.
 * [#202](https://github.com/ruby-grape/grape-entity/pull/202): Fix: Reset `@using_class` memoization on `.setup` - [@rngtng](https://github.com/rngtng).
+* [#203](https://github.com/ruby-grape/grape-entity/pull/203): Fix: return `self` instead of `nil` if no attribute was deleted - [@rngtng](https://github.com/rngtng).
```
dblock: This seems incorrect, shouldn't it say "`Grape::Entity::Exposure::NestingExposure::NestedExposures.delete_if` always
 returns exposures."? It's unlcear what `self` is and what the story here is ;)

rngtng: ah sure.. done

dblock:
Merged, thanks.

rngtng:
np, thank you!

PR #204, opened 2016-01-28
Add option :merge to expose. Allow to merge fields into nested hashes and into the root

### LINKED ISSUE #138: Feature request: Expose a Hash merged to the root, not as a sub key
I'd like to be able to append an arbitrary hash to the final result of the entity, without nesting it.

There are a couple of ways I can think of:

```ruby
class ArticleEntity < Grape::Entity
   expose :title

   # example 1 - most flexible, merge output with this hash
   expose_on_root do |object, options|
      {
         key: "value",
         key_2: object.something
      }
   end

   # example 2: include another entity without nesting it in the result hash
   expose_on_root using: LikeableEntity

   # which is a shortcut for:
   expose_on_root do |object, options|
      LikeableEntity.represent(object, options).as_json
   end
end

class LikeableEntity < Grape::Entity
   expose :like_count
   expose :likers, using: UserEntity
end

ArticleEntity.represent(article).as_json
# => { title: ..., key: ..., key_2: ..., like_count: ..., likers: [...] }
```

Thoughts?

avyy:
This also closes #138
Hi guys. Long time ago I dreamed about merging fields into nested hashes. And that's it. I found it useful and decided to make a PR. 

We can do this:
```ruby
expose :contact_info do
  expose :phone
  expose :address, merge: true, using: API::Entities::Address
end
```

this:
```ruby
expose :something_useful, merge: true
```

And even this:
```ruby
expose :profiles do
  expose :users, merge: true, using: API::Entities::User
  expose :admins, merge: true, using: API::Entities::Admin
end
```

It would be awesome if you take a look at this. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 * Your contribution here.
 
+* [#204](https://github.com/ruby-grape/grape-entity/pull/204): Added ability to merge fields into hashes/root (`:merge` option for `.expose`): [#138](https://github.com/ruby-grape/grape-entity/issues/138) - [@avyy](https://github.com/avyy).
+
 0.5.0 (2015-12-07)
 ==================
 
--- a/README.md
+++ b/README.md
@@ -21,9 +21,10 @@ module API
       expose :text, documentation: { type: "String", desc: "Status update text." }
       expose :ip, if: { type: :full }
       expose :user_type, :user_id, if: lambda { |status, options| status.user.public? }
+      expose :location, merge: true
       expose :contact_info do
         expose :phone
-        expose :address, using: API::Entities::Address
+        expose :address, merge: true, using: API::Entities::Address
       end
       expose :digest do |status, options|
         Digest::MD5.hexdigest status.txt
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 0.5.1 (Next)
 ============
 
-* Your contribution here.
```
dblock: Put this back please.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
...
+* [#204](https://github.com/ruby-grape/grape-entity/pull/204): Added ability to merge fields into hashes/root (`:merge` option for `.expose`). This also closes [#138](https://github.com/ruby-grape/grape-entity/issues/138) - [@avyy](https://github.com/avyy).
```
dblock: Just list the numbers like [#204], [#138]: ... instead of saying "this also closes"...

dblock:
I like it, just amend the CHANGELOG please, and I'll keep this open for a bit, lets see what others say.

uberjay:
Very timely for me! I was just poking around for solutions to this. Specifically, the ability to merge into the root hash. The one question I have, is what's the behavior on key collision, and can I control it. One idea is to support specifying a lambda to merge, and have it work like Hash#merge's block params:

```ruby
expose :custom_attrs, merge: -> (k, v1, v2) { v1 }
```

I *think* this would do the trick -- in the OutputBuilder class:

```ruby
if exposure.for_merge
  if exposure.for_merge.respond_to? :call
    @output_hash.merge! result, &exposure.for_merge
  else
    @output_hash.merge! result
  end
else
  @output_hash[exposure.key] = result
end
```

avyy:
Hi @uberjay .
Default behaviour is just to `merge!` it. So a new key replaces old one. I think to use lambda is a good idea.

dblock:
I am going to merge this.

See what I did there? Merge ...

:+1:

dblock:
@avyy Looks like this introduced a regression, can you please check out https://github.com/ruby-grape/grape-entity/issues/213?

avyy:
@dblock ok I'll check it a bit later :+1:

wleeper:
Any update on this?  Just spent a lot of time chasing my tail due to this.  I recommend pulling 0.5.1 or reverting this until it is resolved.  This is a breaking change for more than just testing.  

It breaks the wrapping of error messages in error! calls when using an entity model because the result returned to the present algorithm is no longer a Hash.

avyy:
@wleeper I working on this. Tomorrow I let you know if it's really better to revert that commit 🎱

dblock:
I don't think yanking 0.5.1 now is a good option, there're lots of people using it in production and it doesn't break "everything". Looking forward to a fix.

lucaskanashiro:
Any news?

dblock:
@lucaskanashiro Looking forward to your PR!

PR #210, opened 2016-02-12
Using the `raise_error` matcher with providing a specific error

u2:
Fix spec warning
>
WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Actual error raised was #<NoMethodError:  missing attribute `nonexistent_attribute' on #[Double (anonymous)]>. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /Users/u2/soft/grape-entity/spec/grape_entity/entity_spec.rb:1136:in `block (5 levels) in <top (required)>'.


dblock:
No need to CHANGELOG for a spec fix, but appreciate you writing it.

Specs now pass without warnings? Add `config.raise_errors_for_deprecations!` into `spec_helper.rb`?

u2:
Just add specific error class or message for the missing one. I don't think suppressing it is a good way.

dblock:
@u2 I mean adding `raise_errors_for_deprecations!` would actually make specs fail loudly instead of producing the WARNING. Without your change if you added `raise_errors_for_deprecations!` the specs should fail, right?

dblock:
Nevermind, I am wrong, it doesn't seem to fail loudly at all, ...

PR #217, opened 2016-04-25
Do not delegate to Kernel methods.

### LINKED ISSUE #215: Kernel methods get priority over delegate methods
Hi,

Question: Due to the way exposures are looked up in the entity class first (calling `respond_to attribute, true`), attributes defined on the delegate class that collide with `Kernel` methods can not be exposed. We today noticed that with a AR model that has a `belongs_to :system` association. Probably poor wording on our side, but still: What would be the preferred way to deal with this? Currently we need to monkey-patch `Grape::Entity#delegate_attribute`.

```
class Foo < ActiveRecord::Base
  belongs_to :system

  class Entity < Grape::Entity
    expose :system
  end
end
```

Thanks,
Malte

maltoe:
This patch prevents Entity from delegating attributes
to Kernel methods. This is accomplished by testing the
methods the Entity class responds to for ownership
by the Entity class itself.

Fixes #215.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.5.2 (Next)
 ============
 
+* [#215](https://github.com/ruby-grape/grape-entity/pull/217): `#delegate_attribute` should not delegate to methods included with `Kernel`. - [@maltoe](https://github.com/maltoe)
 * Your contribution here.
 
 0.5.1 (2016-4-4)
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -7,6 +7,4 @@ require 'bundler'
 
 Bundler.require :default, :test
 
-require 'pry'
-
 RSpec.configure(&:raise_errors_for_deprecations!)
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 language: ruby
 
 cache: bundler
+bundler_args: --without development
 
 rvm:
   - ruby-head
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :development do
 end
 
 group :development, :test do
+  gem 'rake'
   gem 'json'
   gem 'rspec'
   gem 'rack-test', '~> 0.6.2', require: 'rack/test'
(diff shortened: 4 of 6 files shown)


maltoe:
The build error seems unrelated ("listen" gem doesn't work with Ruby < 2.2).

dblock:
This needs a CHANGELOG entry too, please. Try fixing the build too, please, maybe we just need to lock `listen` or whatever brings it down to a version.

maltoe:
There you go. Removed pry from the test suite, too, as imho it doesn't belong there. When you need it to debug, require it manually.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 0.5.2 (Next)
 ============
 
-* Your contribution here.
```
dblock: Put this back below your change entry please.

spec/spec_helper.rb
```diff
@@ -7,6 +7,4 @@
 
 Bundler.require :default, :test
 
-require 'pry'
```
dblock: 👍

dblock:
Btw, https://github.com/ruby-grape/grape is failing the same way, if you feel generous today and want to fix :)

maltoe:
I've put the "Your contrib here" statement back in. Sorry, can't spend any more time on this today for grape, maybe some other time.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 0.5.2 (Next)
 ============
 
+* [#215](https://github.com/ruby-grape/grape-entity/pull/217): `#delegate_attribute` should not delegate to methods included with `Kernel`.
```
dblock: One last nitpick and then I stop. See the format of these below, your name should be at the end, so `- [@...]().`. Thanks.

maltoe:
No big deal, keep the code quality up :smile:

dblock:
I merged. And still am amending the CHANGELOG a bit since the period is in the wrong place :)

dblock:
Thanks for contributing this! It was a major PITA.

maltoe:
You seem to have a bit too much time :) But thanks for merging! looking forward to getting rid of our monkey-patch.

PR #219, opened 2016-05-04
fix-serializable-hash-option

sbatykov:


dblock:
I am not sure what this does, it needs specs and a CHANGELOG entry please?

sbatykov:
`        def serializable_value(entity, options)
          partial_output = valid_value(entity, options)

          if partial_output.respond_to?(:serializable_hash)
            partial_output.serializable_hash
          elsif partial_output.is_a?(Array) && partial_output.all? { |o| o.respond_to?(:serializable_hash) }
            partial_output.map(&:serializable_hash)
          elsif partial_output.is_a?(Hash)
            partial_output.each do |key, value|
              partial_output[key] = value.serializable_hash if value.respond_to?(:serializable_hash)
            end
          else
            partial_output
          end
        end
`
Options already pass to output in **partial_output = valid_value(entity, options)**
and can be modified self-own way.
When call **partial_output.serializable_hash(option)** options will be merged. 

Also all others else branch, didn't pass **option ** to **serializable_hash**. May be it's old code.

I wall add CHANGLOG entry.

dblock:
Ok makes sense. You need to write a test that shows a failure without this fix, please.

sbatykov:
commit test and change log

spec/grape_entity/hash_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+describe Grape::Entity::Hash do
+  let(:fresh_class) { Class.new(Grape::Entity) }
+
+  it 'Test' do
+    class Address < Grape::Entity
```
dblock: These are global scope. Can you declare them with `let(:address) { Class.new(...) }` please?

spec/grape_entity/hash_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    entity = { post: '123456',
+               city: 'city',
+               street: 'street',
+               house: 'house',
+               something_else: 'something_else' }
+
+    expect(Address.represent(entity).serializable_hash).to eq \
+      entity.slice(:city, :street, :house)
+    expect(Address.represent(entity, full: true).serializable_hash).to eq \
+      entity.slice(:post, :city, :street, :house)
```
dblock: Maybe break this up with separate tests?

dblock:
Thanks. See my comments on tests. Squash your commits too please if you can.

dblock:
Merged.

PR #222, opened 2016-05-23
Update README.md

glaucocustodio:
I think nice to show we can reuse formatters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,16 @@ with_options(format_with: :iso_timestamp) do
 end
 ```
 
+```ruby
+# Helper module for instance
+Grape::Entity.format_with :utc do |date|
+  date.utc if date
+end
+
+# Entity
+expose :updated_at, format_with: :utc
+```
+
 #### Documentation
 
 Expose documentation with the field. Gets bubbled up when used with Grape and various API documentation systems.
--- a/README.md
+++ b/README.md
@@ -311,21 +311,42 @@ expose :replies, using: API::Entities::Status, as: :responses
 Apply a formatter before exposing a value.
 
 ```ruby
-format_with(:iso_timestamp) { |dt| dt.iso8601 }
-with_options(format_with: :iso_timestamp) do
-  expose :created_at
-  expose :updated_at
+module Entities
+  class MyModel < Grape::Entity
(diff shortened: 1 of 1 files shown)


dblock:
Thanks. Maybe combine it with the example above and add some explanations in plain English?

glaucocustodio:
How about that @dblock:

Apply a formatter before exposing a value.

```ruby
format_with(:iso_timestamp) { |dt| dt.iso8601 }
with_options(format_with: :iso_timestamp) do
  expose :created_at
  expose :updated_at
end
```

Defining a reusable formatter between multiples entities:

```ruby
# Helper module for instance
Grape::Entity.format_with :utc do |date|
  date.utc if date
end
 
# Entity
expose :updated_at, format_with: :utc
```

dblock:
The language is fine, but I'd like the two code examples to be consistent, for example the first doesn't have `Grape::Entity.` and the second does. One uses a block with `do` and the other one has `{`.

glaucocustodio:
Just `format_with` inside a helper raises an exception like `undefined method format_with for ApiHelpers:Module`, that is the reason for `Grape::Entity.` before. Maybe we can give a bit more of context like below. `{` is the proper style for inline blocks, we can change to a multi line block if necessary (I don't think so).

Apply a formatter before exposing a value.

```ruby
module Entities
  class MyModel < Grape::Entity
    format_with(:iso_timestamp) { |dt| dt.iso8601 }

    with_options(format_with: :iso_timestamp) do
      expose :created_at
      expose :updated_at
    end    
  end
end
```

Defining a reusable formatter between multiples entities:

```ruby
module ApiHelpers
  extend Grape::API::Helpers

  Grape::Entity.format_with :utc do |date|
    date.utc if date
  end
end
```

```ruby
module Entities
  class MyModel < Grape::Entity
    expose :updated_at, format_with: :utc
  end

  class AnotherModel < Grape::Entity
    expose :created_at, format_with: :utc
  end
end
```

dblock:
Much better, amend and I'll merge.

dblock:
I still have major ADD around `{` vs. `do` btw!

glaucocustodio:
Please, check now.

dblock:
Looks great, merging.

PR #225, opened 2016-06-02
Update README.md

glaucocustodio:


dblock:
This is a monkey patch and I don't think we should be documenting monkey-patches. I don't recommend just adding methods to `Grape::Entity` unless we think they actually belong in it (aka legitimate extensions that we want to have as features). 

I'm going to close this without merging, but maybe you want to describe the actual scenario you're trying to solve and I'll help you write it a bit cleaner?

glaucocustodio:
When calling `format_with:` with a block or `using:` option, the formatter is ignored:
```ruby
# formatter `foo` defined in a helper
Grape::Entity.format_with :foo do |value|
  binding.pry
end

# formatter `foo` is not called, totally ignored
expose :users, using: Entities::MyEntity, format_with: :foo do |instance, _opt|
  instance.users
end

# formatter `foo` is not called, totally ignored
expose :users, using: Entities::MyEntity, format_with: :foo

# only this works
expose :users, format_with: :foo
```

I can even call a non-existing formatter and any error is raised:

```ruby
# formatter `bar` does not exist, again, totally ignored
expose :users, using: Entities::MyEntity, format_with: :bar do |instance, _opt|
  instance.users
end

# formatter `bar` does not exist, again, totally ignored
expose :users, using: Entities:: MyEntity, format_with: :bar

# only this raises the error: 
# NoMethodError: undefined method `bar' for #<Entities...
expose :users, format_with: :bar
```

The way I found to format in the cases above where the formatter is ignored is to mokey-patching `Entity`.

dblock:
@glaucocustodio That's another bug? Open a separate issue please.

glaucocustodio:
I found out we in fact cannot use both `using`, `proc`, `format_with` and `nesting` together as coded https://github.com/ruby-grape/grape-entity/blob/master/lib/grape_entity/exposure.rb#L17.

glaucocustodio:
It's seems not to be a bug, but I think it is a nice feature to have.. Should I open a issue with this feature request?

dblock:
Sure. Or better implement it and make a pull request.

PR #226, opened 2016-06-22
Add fetch method to fetch from opts_hash

alanjcfs:
We use `fetch` often to get options (raising an error if an option doesn't exist) as well as to provide a alternative value if an option is not provided.


dblock:
I like it. This needs tests and a CHANGELOG entry please.

alanjcfs:
@dblock Thanks for the feedback. I added the specs at the end of the spec as there was no specific specs for Grape::Entity::Options. Let me know if there's anything to improve. I also squashed the commits.

dblock:
Merged, looks good.

reneklacan:
It's pretty long time so maybe it would be nice to release new version with this 😇

dblock:
@idyll @marshall-lee ^^^

PR #227, opened 2016-06-23
Expose Hash objects with string keys

### LINKED ISSUE #223: Grape::Entity::Delegator::HashObject fails when extracting string keys
Hi,

I was trying to present a hash of values with a grape entity. The keys in the hash were strings and I got null values all over the place as a response.

It boils down to Grape::Entity::Delegator::HashObject, when it tries to render the attribute it only tries to extract the attribute with a symbol as a key and not a string. 

screen shot 2016-05-30 at 13 50 29

I think a simple 
`object[attribute] || object[attributeto_s]`
would fix the issue, although I'm not sure whether you need to make the destinction between symbols and strings there for other reasons. For now I simple convert all my keys to symbols before handing them to grape entity. But in any case would be more convenient the other way around.

Cheers 
Tom

nbulaj:
Fix #223: extracting of string keys for `Grape::Entity::Delegator::HashObject`. Now Hashes with string keys (database records with JSON fields for example) can be exposed correctly.

### DIFF
--- a/lib/grape_entity/delegator/hash_object.rb
+++ b/lib/grape_entity/delegator/hash_object.rb
@@ -3,7 +3,7 @@ module Grape
     module Delegator
       class HashObject < Base
         def delegate(attribute)
-          object[attribute]
+          object[attribute] || object[attribute.to_s]
         end
       end
     end
--- a/spec/grape_entity/entity_spec.rb
+++ b/spec/grape_entity/entity_spec.rb
@@ -626,6 +626,12 @@ describe Grape::Entity do
         expect(representation).to eq(awesome: true)
       end
 
+      it 'returns a serialized hash of a hash with string keys' do
+        subject.expose('awesome', 'data')
+        representation = subject.represent({ 'awesome' => true, 'data' => 'awesome' }, serializable: true)
+        expect(representation).to eq(awesome: true, data: 'awesome')
+      end
+
       it 'returns a serialized hash of an OpenStruct' do
         subject.expose(:awesome)
         representation = subject.represent(OpenStruct.new, serializable: true)


dblock:
I am not 100% sure we want this. After all keys of different types aren't equivalent, so you can very well have two values, one with `:key` and another with `"key"`. Thoughts?

This needs a CHANGELOG entry, too.

nbulaj:
@dblock no problem with different keys (small change of the `hash_spec.rb`):

```ruby
require 'spec_helper'

describe Grape::Entity do
  it 'except option for nested entity' do
    module EntitySpec
      class Address < Grape::Entity
        expose :post, if: :full
        expose :city
        expose :street
        expose :house
      end

      class Company < Grape::Entity
        expose :full_name, if: :full
        expose :name
        expose :address do |c, o|
          Address.represent c[:address], Grape::Entity::Options.new(o.opts_hash.except(:full))
        end
      end
    end

   # Mixed object to be exposed
    company = {
      'full_name' => 'full_name',
      'name' => 'name',
      address: {
        post: '123456',
        city: 'city',
        'street' => 'street',
        house: 'house',
        something_else: 'something_else'
      }
    }

    # Output result without option :full
    short_company = {
      name: 'name',
      address: {
        city: 'city',
        street:'street',
        house: 'house'
      }
    }

   # Output result with :full option
    full_company = {
      full_name: 'full_name',
      name: 'name',
      address: {
        city: 'city',
        street:'street',
        house: 'house'
      }
    }

    expect(EntitySpec::Company.represent(company).serializable_hash).to eq short_company
    expect(EntitySpec::Company.represent(company, full: true).serializable_hash).to eq full_company
  end
end

```

dblock:
How do I expose both `:city` and `'city'`? I understand this is not a real scenario, but there's now an automatic "i don't care whether your key is a symbol or string" assumption, and in the "symbol then string" order, which is a lot IMO.

akozhemiakin:
I'm not very familiar with Grape Entity implementation details, so it may be impossible to make things work that way but...
Maybe it is possible let users specify either string or symbol when they expose things? Like that:
```ruby
class MyEntity < Grape::Entity
  expose 'label' # exposing object['label']
  expose :quantity # exposing object[:quantity]
end
```

Also, if it is possible, than it may be useful to introduce some fallback strategy.
```ruby
class MyEntity < Grape::Entity
  # exposing object['label'] or object[:label] if object['label'] == nil
  expose 'label'

  # exposing object[:quantity] or object['quantity' if object[:quantity] == nil
  expose :quantity
end
```

dblock:
I believe this is how it works today @akozhemiakin, and we (someone) should add tests for that behavior - this specific PR would be a no-go. I think to solve the original problem in this PR I would try adding a mixin like `Hashie::Extensions::IndifferentAccess ` instead of hacking grape-entity.

nbulaj:
> I believe this is how it works today

No, it doesn't work like so.

```ruby
    module EntitySpec
      class Address < Grape::Entity
        expose 'city'
        expose 'street'
        expose 'house'
      end

      class Company < Grape::Entity
        expose 'name'
        expose 'address' do |c, _|
          Address.represent c['address']
        end
      end
    end

    company = {
      'name' => 'name',
      'address' => {
        'city' => 'city',
        'street' => 'street',
        'house' => 'house',
      }
    }

    puts EntitySpec::Company.represent(company).serializable_hash.inspect

    #=> {:name=>nil, :address=>{:city=>nil, :street=>nil, :house=>nil}}
```

>  I would try adding a mixin like Hashie::Extensions::IndifferentAccess

I have not heard of it before, but I think it would be a great solution!

dblock:
Ok I am with you @nbulaj. Can you try with the Hashie extension first? I just think it's a better solution if that works than hacking here.

dblock:
Also try `company` above as a http://apidock.com/rails/ActiveSupport/HashWithIndifferentAccess, does that work out of the box?

nbulaj:
@dblock try to change `hash_spec.rb` making `company` a `HashWithIndifferentAccess`, and it works. Well, specs are broken, but the result of the `.serializable_hash` is exactly what we expect (`{:name=>"name", :address=>{:city=>"city", :street=>"street", :house=>"house"}}
`).

```ruby
 expected: {"name"=>"name", :address=>{:city=>"city", :street=>"street", :house=>"house"}}
            got: #<Grape::Entity::Exposure::NestingExposure::OutputBuilder({:name=>"name", :address=>#<Grape::Entity::Exposure::NestingExposure::OutputBuilder({:city=>"city", :street=>"street", :house=>"house"})>})>
     
       (compared using ==)
     
       Diff:
       @@ -1,3 +1,2 @@
       -"name" => "name",
       -:address => {:city=>"city", :street=>"street", :house=>"house"},
       +{:name=>"name", :address=>{:city=>"city", :street=>"street", :house=>"house"}}

```

If you need, I can add a modified spec code.

So, it works only with `Hash`es with indifferent access and everybody need to make them as such.

dblock:
Ok, what should I make of this PR? Close it?

akozhemiakin:
For me it does not look very convenient. That way we will be forced to convert regular hashes to hashes with indifferent access every time we need to expose them. Not a lot, but still kind of boilerplate.

nbulaj:
I agree with @akozhemiakin . We must do the best choise

LeFnord:
it seems there is no progress, so I'll close it

Issue #213, opened 2016-04-06
0.5.1 broke Grape specs


dblock:
See https://github.com/ruby-grape/grape/issues/1350, possibly a regression?

dblock:
Bisected this down to https://github.com/ruby-grape/grape-entity/pull/204.

avyy:
I will be able to check it at the end of week guys

avyy:
@dblock solved?

dblock:
That's really a question for https://github.com/ruby-grape/grape/issues/1350, update *grape* to use 0.5.1, see what happens.

PR #232, opened 2016-06-30
Let the OutputBuilder be the Hash as well

avyy:
#213 
Hi everyone :suspect: 
The problem that we've changed the `Hash` output to the `OutputBuilder` one (see #204) . Even though an `OutputBuilder` instance delegates everything to its `output` object (which is a `Hash`) - it  returns `false` when we do this: `.is_a?(Hash)` on a `OutputBuilder` instance. `is_a?` method is not delegated. And grape errors formatters expect to get a `Hash` instance from `Entity` after representation, not `OutputBuilder` one.

I checked grape master branch and my grape-entity branch tests. Everything should be fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * [#215](https://github.com/ruby-grape/grape-entity/pull/217): Fix: `#delegate_attribute` no longer delegates to methods included with `Kernel` - [@maltoe](https://github.com/maltoe).
 * [#219](https://github.com/ruby-grape/grape-entity/pull/219): Fix: double pass options in serializable_hash - [@sbatykov](https://github.com/sbatykov).
 * [#226](https://github.com/ruby-grape/grape-entity/pull/226): Add fetch method to fetch from opts_hash - [@alanjcfs](https://github.com/alanjcfs).
+* [#232](https://github.com/ruby-grape/grape-entity/pull/232): Add `#kind_of?` and `#is_a?` to `OutputBuilder` to get an exact class of an `output` object. [#213](https://github.com/ruby-grape/grape-entity/issues/213) - [@avyy](https://github.com/avyy).
 * Your contribution here.
 
 0.5.1 (2016-4-4)
--- a/grape-entity.gemspec
+++ b/grape-entity.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
 
   s.rubyforge_project = 'grape-entity'
 
-  s.add_runtime_dependency 'activesupport'
+  s.add_runtime_dependency 'activesupport', '< 5'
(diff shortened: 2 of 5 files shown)


dblock:
The build is toast, but I don't think it's your fault. If you don't mind taking a look though please.

avyy:
MRI 2.3: ✅ 
Lower ruby versions: `Gem::InstallError: activesupport requires Ruby version >= 2.2.2.`. I think it's happened because of `activesupport (5.0.0)` which was released yesterday. We can add `<= 4.2.6` to gemspec for activesupport.

dblock:
Yes please

avyy:
The same problem with `rack` and `json`. Looks good now.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * [#215](https://github.com/ruby-grape/grape-entity/pull/217): Fix: `#delegate_attribute` no longer delegates to methods included with `Kernel` - [@maltoe](https://github.com/maltoe).
 * [#219](https://github.com/ruby-grape/grape-entity/pull/219): Fix: double pass options in serializable_hash - [@sbatykov](https://github.com/sbatykov).
 * [#226](https://github.com/ruby-grape/grape-entity/pull/226): Add fetch method to fetch from opts_hash - [@alanjcfs](https://github.com/alanjcfs).
+* [#232](https://github.com/ruby-grape/grape-entity/pull/232): Fix: grape tests are green [#213](https://github.com/ruby-grape/grape-entity/issues/213) - [@avyy](https://github.com/avyy).
```
dblock: This isn't material for users, so remove it.

dblock: However, the `kind_of` change is relevant, so you should say something to that regard.

grape-entity.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
 
   s.rubyforge_project = 'grape-entity'
 
-  s.add_runtime_dependency 'activesupport'
+  s.add_runtime_dependency 'activesupport', '<= 4.2.6'
```
dblock: I think `< 5` is the right entry here. I am however worried that we're effectively saying we're not compatible with activesupport 5, while the code is and the problems are structural. Maybe we should just fix things around here to work with AS 5, like increment the minimum Ruby version?

avyy: @dblock  We have to make a choice what is more important: supporting old rubies or supporting last AS/rack/json.

dblock: I'll take this change as is for now, and we (you) should open an issue to support latest AS. 

Can you amend the changelog entry please and I'll merge?

avyy:
@dblock check last changes pls

dblock:
Merging, thanks.

Issue #233, opened 2016-07-04
Latest ruby version support and dependencies


avyy:
We did not specified `active_support, rack, json` versions in `gemspec` and `Gemfile`, so last update of those gems broke our travis [builds](https://travis-ci.org/ruby-grape/grape-entity/builds/141552591), because those gems require `ruby >= 2.2.2` and travis configured to run specs on ruby `1.9.3 .. 2.3`.

We specified needed versions of dependencies to pass travis' tests on all configured platforms temporarily [(see gemspec and Gemfile)](https://github.com/ruby-grape/grape-entity/commit/c5abc1bc83ee01f8cb23b4e77e21846bcb733157). But we have to make a choice what to do next: use latest versions of dependencies or support old rubies.

What do you think? :suspect:

dblock:
I am not in love with the idea of requiring Ruby 2.2 on HEAD for all Grape projects just because ActiveSupport 5 does. I suggest trying to condition the dependencies in `Gemfile` and remove any version locking in `.gemspec` next. Something like this:

Gemfile:

```ruby
if RUBY_VERSION ...
gem 'activesupport', '~> 4...'
elsif ...
gem 'activesupport', '~> 5...'
end
```

What do you think?

nbulaj:
You can use different gemfiles and test them against different versions of gems, like it done here: [doorkeeper-sequel/blob/master/.travis.yml](https://github.com/nbulaj/doorkeeper-sequel/blob/master/.travis.yml).

dblock:
Looking forward to some PRs!

PR #234, opened 2016-07-07
Add ruby version checking in Gemfile to install needed gems versions

avyy:
#233 
Can we improve this?


dblock:
I like it. Personally I prefer much "dumber" approaches, but this works well. Merging.

PR #231, opened 2016-06-28
Fix for #215. Allow delegate_attribute for derived classes

sbatykov:

PR #238, opened 2016-09-08
Allow methods from ancestors of the entity

anakinj:
Had a little problem with inherited entities and changes done in #217. This should allow the method to be in a inherited class up until the Grape::Entity class.


anakinj:
Just noticed that this was fixed in #231 already. So will close this, also the approach in #231 is nicer:).

PR #248, opened 2016-11-15
Fix nil values causing errors when merge option passed

arempe93:
Currently, if something is exposed with `merge: true` and the value ends up being `nil`, it raises an unhandled exception. This changes that behavior to add nothing to the output and throw no exception

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


LeFnord:
thanks @arempe93 … please can you make danger happy (copy it  over)

arempe93:
sorry - could you explain what copy it over is? i thought I had fixed the danger problem with the changelog

LeFnord:
all fine now … 👍

PR #250, opened 2016-11-20
Fix #249: leaking of options and internals in default serialization.

### LINKED ISSUE #249: Grape::Entity leaks @options and other internals when presented without a formatter
When presented without a formatter a Grape::Entity gets serialized with all its internals, including `@options`. This doesn't appear to leak any sensitive information most of the time, but YMMV as the rack app, session, request objects and ActiveRecord internals come along.

It's a nasty side effect of adding something like `content_type :csv, 'text/csv'` and not having a general `formatter :csv, ...` that catches everything and calls, for example, `to_csv`. Entities by default support `to_xml` and `to_json`, but nothing else.

Complete example in https://gist.github.com/dblock/5b49741468efd5b6a7d6828167d47a65.

dblock:
I think the best option is to make `inspect` serialize the object.


dblock:
@idyll @marshall-lee @LeFnord would appreciate any +/- 1 comments

LeFnord:
0.6.0 release ?

dblock:
@LeFnord done.

PR #251, opened 2016-11-25
Avoid noise when code runs with Ruby warnings

cpetschnig:
New versions of RSpec have warnings turned on by default:

    # This setting enables warnings. It's recommended, but in some cases may
    # be too noisy due to issues in dependencies.
    config.warnings = true

I believe these warnings are helpful in general to write better code. However, Grape Entity is very noisy. I get more than 10 lines of warnings for one single test where Grape Entity is involved.

I hope you agree. Some of my changes could also be solved differently for sure. I am happy to discuss and change those cases. I also turned warnings on in the test suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 #### Features
 
-* Your contrbution here.
+* Your contribution here.
 
 #### Fixes
 
-* Your contrbution here.
+* [#251](https://github.com/ruby-grape/grape-entity/pull/251): Avoid noise when code runs with Ruby warnings - [@cpetschnig](https://github.com/cpetschnig).
+* Your contribution here.
 
 ### 0.6.0 (2016-11-20)
 
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -13,7 +13,7 @@ Metrics/AbcSize:
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 206
+  Max: 208
 
 # Offense count: 3
 Metrics/CyclomaticComplexity:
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,11 @@
 
 #### Features
 
-* Your contrbution here.
+* Your contribution here.
```
LeFnord: thanks

CHANGELOG.md
```diff
@@ -2,11 +2,11 @@
...
 
 #### Fixes
 
-* Your contrbution here.
```
LeFnord: please re-add it below your contribution

LeFnord:
thanks @cpetschnig for contributing, please see my comment above

spec/spec_helper.rb
```diff
@@ -7,4 +7,8 @@
 
 Bundler.require :default, :test
 
-RSpec.configure(&:raise_errors_for_deprecations!)
+RSpec.configure do |config|
+  config.raise_errors_for_deprecations!
+
+  config.warnings = true
+end
```
LeFnord: please revert this change

LeFnord: @dblock this would enable the warnings and makes also the specs noisy, maybe they could be used for writing better code, but think for that we have rubocop, so by enabling it, we have two of such tools … mmh

if they both give the same or similar warnings, then one is enough and the rubocop messages much more helpful, IMHO the rspec warnings are higher level and helpful for experienced devs, but they also raises the barrier for contributions

if not then it should be disabled

dblock: I agree with @LeFnord.

cpetschnig:
Thank you for your feedback. I have updated the PR.

LeFnord:
👍

PR #264, opened 2017-04-05
Rubocop cleanup

james2m:
Adds a Rubocop config for 0.48.1 based on the config from grape,  and update the todo list for the existing offences. I don't have time to fix all those issues so I suggest tackling the todo should be a separate PR.


.rubocop.yml
```diff
@@ -1,6 +1,30 @@
 AllCops:
+  TargetRubyVersion: 2.1
```
LeFnord: please change it to 2.4, maybe alos in the grape

LeFnord:
please run `bundle exec rubocop -a` to fix the frozen string luteral stuff, thanks

LeFnord:
thanks @james2m

dblock:
FWIW I generally just stick to `rubocop -a ; rubocop --auto-gen-config`. It's easier to think about.

PR #268, opened 2017-04-18
Loosen the activesupport dependency

anakinj:


grape-entity.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.rubyforge_project = 'grape-entity'
 
   s.add_runtime_dependency 'multi_json', '>= 1.3.2'
-  s.add_runtime_dependency 'activesupport', '>= 5.0.0'
+  s.add_runtime_dependency 'activesupport'
```
LeFnord: please change instead to: `> 4`, else it would fail this spec`./spec/grape_entity/hash_spec.rb:6 # Grape::Entity except option for nested entity`) with `< 4`

anakinj:
I still changed this version to >=4.0.0 to be more schematically pretty, and allow the 4.0.0 version of activesupport :)

PR #272, opened 2017-07-05
Relax minimal Ruby version requirement

joneslee85:
Ruby 2.2 is still alive and fixing minimal version to 2.2.6 would cause many issues for applications out there that are still on older version of series 2.2

### DIFF
--- a/grape-entity.gemspec
+++ b/grape-entity.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.description = 'Extracted from Grape, A Ruby framework for rapid API development with great conventions.'
   s.license     = 'MIT'
 
-  s.required_ruby_version = '>= 2.2.6'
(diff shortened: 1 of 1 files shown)


dblock:
I think we do want a CHANGELOG for this, something like "allow any Ruby version >= 2.2"

LeFnord:
IMHO, I would prefer to update the ruby version to the more secure and stable one (e.g. 2.2.7)  to avoid issues, instaed to use the gem with an untested one → specs run actual for 2.2.7

dblock:
@LeFnord I think we cannot force people into it though, this library is compatible with 2.2.whatever so we shouldn't force imply otherwise.

That said I want to know for which nuclear power plant @joneslee85 works for :)

LeFnord:
thanks @joneslee85

PR #282, opened 2017-11-01
Use latest ruby version in the major versions

kachick:


LeFnord:
thanks @kachick

kachick:
Thank you!

PR #288, opened 2017-12-28
Fix wrong argument exception when &:block passed to the expose method

DmitryTsepelev:
I've faced this problem when Rubocop suggested to fix `Style/SymbolProc` problem in the following code:

```ruby
expose :something do |object|
  object.any_method
end
```

After the fix the following code started to throw wrong argument number exception, because options are always passed to the block:
```ruby
expose :something, &:any_method
```

I'm aware of the `as:` option which solves the case above, but probably this enhancement will help avoid confusion with Rubocop users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


DmitryTsepelev:
@LeFnord could you please review when you have a second? thanks :)

LeFnord:
the rubocop warnings will be fixed in #287

DmitryTsepelev:
@LeFnord Got it, thank you 👍

PR #290, opened 2018-01-08
Loosens ActiveSupport to >= 3.0.0. Fixes #289

### LINKED ISSUE #289: Loosening ActiveSupport version to 3?
I currently am using Rails 3.2 and would like to use grape-entity. Is there any reason why the Activesupport version couldn't be loosened to 3? I know it was loosened to 4 in the past. 

If this would be alright, I'd be happy to do any sort of testing necessary. The unit tests pass for me on with `'activesupport', '>= 3.0.0'` so I'm not sure if there's any other testing required.

ericschultz:
All the tests pass. `bundle exec rake` has some rubocop reports but they're not related to the change.

### DIFF
--- a/grape-entity.gemspec
+++ b/grape-entity.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.rubyforge_project = 'grape-entity'
 
-  s.add_runtime_dependency 'activesupport', '>= 4.0.0'
+  s.add_runtime_dependency 'activesupport', '>= 3.0.0'
   s.add_runtime_dependency 'multi_json', '>= 1.3.2'
 
   s.add_development_dependency 'bundler'


LeFnord:
I tested it with `'activesupport', '~>3.0'` → uses 3.2.22.5, there one error occurs …
```
 1) Grape::Entity except option for nested entity
     Failure/Error: Address.represent c[:address], Grape::Entity::Options.new(o.opts_hash.except(:full))
     
     NoMethodError:
       undefined method `except' for {:attr_path=>[:address]}:Hash
     # ./spec/grape_entity/hash_spec.rb:19:in `block in <class:Company>'
     ...
     # ./lib/grape_entity/entity.rb:490:in `serializable_hash'
     # ./spec/grape_entity/hash_spec.rb:36:in `block (2 levels) in <top (required)>'
```

it seems, that if the people will really use version 3.x of it, then it didn't work correct

LeFnord:
… so I'll close it

ericschultz:
Sorry about that.

Just looking into this, the problem seems to be that activesupport's extensions for the hash class aren't being included in the spec suite. When I add `require 'active_support/core_ext/hash'` to the top of spec_helper.rb, the specs pass on `'activesupport', '~>3.0'` I'm not sure on the best way to address this (I'm still a bit of a beginner to Ruby).

I'm happy to correct the problem if you have any suggestions.

LeFnord:
ok, then it is so, I'll re-open the issue

Issue #286, opened 2017-12-01
expose as: in child class doesn't override parent class exposure


magni-:
```ruby
class BaseEntity < Grape::Entity
  expose :id
end

class FooEntity < BaseEntity
  expose :id, as: :foo_id
end
```

This used to work in 0.4.5, but stopped once we upgraded.

Whereas before we'd have `{foo_id: 1}`, we now get `{id: 1, foo_id: 1}`

edit: the temp fix is to add `unexpose :id` to `FooEntity`, but that's a bit ugly... and it used to work without it :)

PR #292, opened 2018-01-15
Allow replace non-conditional non-nesting exposures in child classes

DmitryTsepelev:
Tried to fix issue #286 by allowing to override exposures in child classes when the exposure is not conditional and not nesting


LeFnord:
thanks @DmitryTsepelev … please make rubocop happy, and I'm fine to mörge it

DmitryTsepelev:
@LeFnord done, thanks

PR #295, opened 2018-01-25
ver 0.7.0 breaks multiple exposures using same element

jnardone:


LeFnord:
solved with [!298](https://github.com/ruby-grape/grape-entity/pull/298)

Issue #296, opened 2018-01-25
Can't expose same field more than once


jnardone:
ref: https://github.com/ruby-grape/grape-entity/pull/295

In our code we expose the same field more than once, using different formatters e.g.

```
    class Entity < Grape::Entity
        expose :useragent, format_with: :ua_to_os, safe: true, as: :os
        expose :useragent, format_with: :ua_to_browser, safe: true, as: :browser
    end
```

(these `format_with` formatters call some helper methods on some 3rd party code to extract the appropriate bits)  The formatters don't matter though - as the UT in the pull request shows, any use of the field more than once exhibits the problem.

However, starting in 0.7.0, only the last exposure for a given field is represented - all the others are now missing from the output.  I think (but am not sure) that @DmitryTsepelev's changes in https://github.com/ruby-grape/grape-entity/pull/292 might be the culprit. 

Note that the example above (and in the PR as well), there is no class hierarchy - all of these inherit direct from `Grape::Entity`.

DmitryTsepelev:
@jnardone let me take a look

DmitryTsepelev:
@jnardone Okay, investigated a bit and found out that according to the https://github.com/ruby-grape/grape-entity/issues/286 which I was trying to fix we wanted to allow overriding condition in the inherited models (see the example in the issue). However, in your case, we want to allow multiple exposures with the different keys exposed from the same model.
I see two possible solutions here:
1. Improve my condition from the previous PR in a way that it knows where the exposure comes from and, if the class is the same - do not replace it with the next one, if it's a child class - replace according to the other conditions.
2. Allow setting up a special key (e.g. `override_parent`) in the child class and set the default to `false`

cc @LeFnord thoughts?

DmitryTsepelev:
Example implementation of 2 is here https://github.com/DmitryTsepelev/grape-entity/commit/ac8bf226df5a3250f09c2c443ba1b65644d886d1 please let me know if you're ok with that and I'll make a PR shortly

jnardone:
An alternate idea would be - for your #286 fix, only consider it to be a duplicate if the field AND the formatter AND the `as` are unique?  So

```
expose :a
expose :a as :z
```

would only expose `z` but 

```
expose :a, format_with: :foo, as: :a1
expose :a, format_with: :bar, as: a2
expose :a
```

would output all 3 exposures, because they contain distinct modifiers?

DmitryTsepelev:
[...]

PR #297, opened 2018-01-29
Introduce override option for expose

DmitryTsepelev:
Per discussion here https://github.com/ruby-grape/grape-entity/issues/296. Please note that I've removed all other conditions, so even conditional or nesting exposure can be overwritten (should I throw an error or something when people configure their exposures in strange ways?)

### DIFF
--- a/spec/grape_entity/entity_spec.rb
+++ b/spec/grape_entity/entity_spec.rb
@@ -477,10 +477,19 @@ describe Grape::Entity do
           expect(child_class.represent({ name: 'bar' }, serializable: true)).to eq(email: nil, name: 'foo')
         end
 
-        it 'overrides parent class exposure' do
+        it 'not overrides exposure by default' do
           subject.expose :name
           child_class = Class.new(subject)
           child_class.expose :name, as: :child_name
+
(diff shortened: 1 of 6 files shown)


spec/grape_entity/entity_spec.rb
```diff
@@ -477,11 +477,20 @@ class Parent < Person
           expect(child_class.represent({ name: 'bar' }, serializable: true)).to eq(email: nil, name: 'foo')
         end
 
-        it 'overrides parent class exposure' do
+        it 'not overrides exposure by default' do
```
LeFnord: did we have an example which overrides the exposure?

DmitryTsepelev: Yep, I've used the spec I've written before https://github.com/ruby-grape/grape-entity/pull/297/files/0e86ebb5302ed9d2d4effc4d92eda7940abd13e1#diff-2f09e9dc67c5575f0172b3d0defe0e1bR489

LeFnord: ok … thanks for clearing

LeFnord:
Thanks for you work @DmitryTsepelev

PR #300, opened 2018-03-28
Loosens ActiveSupport to 3

ericschultz:
This is a fix for #289 and a second attempt (original pull request at #290)

The spec_helper now loads the needed hash extensions for ActiveSupport < 4. In ActiveSupport >=4, there's no change other than the ActiveSupport version info being loaded, which has no effect test functioning.


spec/spec_helper.rb
```diff
@@ -3,6 +3,12 @@
 require 'simplecov'
 require 'coveralls'
 
+# This works around the hash extensions not being automatically included in ActiveSupport < 4
+require 'active_support/version'
+require 'active_support/core_ext/hash' if ActiveSupport::VERSION &&
+                                          ActiveSupport::VERSION::MAJOR &&
+                                          ActiveSupport::VERSION::MAJOR < 4
```
LeFnord: is `ActiveSupport.version.to_s < '4'` the same?

grape-entity.gemspec
```diff
@@ -18,8 +18,7 @@ Gem::Specification.new do |s|
 
   s.rubyforge_project = 'grape-entity'
 
-  s.add_runtime_dependency 'activesupport', '>=4.0'
-  # FIXME: remove dependecy
+  s.add_runtime_dependency 'activesupport', '>= 3.0.0'
```
LeFnord: please can you change the lines, `FIXME` should be above `multi_json`, thanks

spec/spec_helper.rb
ericschultz: ActiveSupport 3 has a different API for version (https://github.com/rails/rails/blob/3-2-stable/activesupport/lib/active_support/version.rb). It's closer to 'activesupport/gem_version' in 4+.

grape-entity.gemspec
ericschultz: Will do.

LeFnord:
thanks @ericschultz …

will rebase and mörge it, after this one https://github.com/ruby-grape/grape-entity/pull/301, ok

Issue #258, opened 2017-02-13
Exposure doesn't call method that was included through a module


yilangmok-gusto:
We have a module that includes common functionality in multiple entities.
```
module MyModule
  def a_value
    ...
  end
end

class A < Grape::Entity
  include MyModule
  expose :a_value
end
```
As of #215 and #231, this no longer works because `a_value`'s owner is not a `Grape::Entity` according to this code:
```
def delegate_attribute(attribute)
  if respond_to?(attribute, true) && Grape::Entity > method(attribute).owner
    send(attribute)
  else
    ...
  end
end
```
Any ideas how we can work around this? Thanks!

PR #307, opened 2018-09-04
Allow exposures to call methods defined in included modules

robertoz-01:
* Fix https://github.com/ruby-grape/grape-entity/issues/258

It makes possible for the exposure to use methods defined in modules included in the entity. eg:

```
module MyModule
  def a_value
    ...
  end
end

class A < Grape::Entity
  include MyModule
  expose :a_value
end
```


LeFnord:
thanks @robertoz-01 … and sorry for the late reply … think this is a welcome feature :)

PR #310, opened 2018-10-13
support multiple merge fields

fangzhu19880123:
I want to use like this
```ruby
expose :config1, merge: true, using: ApplicationEntity
expose :config2, merge: true, using: ApplicationEntity
```
but I must use like this
```ruby
expose :config1, merge: true, using: ApplicationEntity
expose :config2, merge: ->(k, v1, v2) { v2 }, using: ApplicationEntity
```


LeFnord:
thanks @fangzhu19880123 … but please add a spec

PR #315, opened 2019-01-29
CI: Define matrix one way

olleolleolle:
- this keeps CI matrix creation in "rvm" settings
  - trust RVM to have updated RubyGems

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,5 @@ matrix:
   fast_finish: true
 
   allow_failures:
-    - rvm: 2.3.8
     - rvm: ruby-head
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,29 +1,22 @@
...
 matrix:
   fast_finish: true
 
-  include:
-    - rvm: 2.3.8
-    - rvm: ruby-head
-    - rvm: jruby-head
-
   allow_failures:
-    - rvm: 2.3.8
```
LeFnord: this should be re-added … official supporting last 3 versions is enough

PR #319, opened 2019-05-15
Support hashes with string keys

mhenrixon:
Before this change the only way I could find to access a hash with string keys
was to either convert it to a symbol hash or define methods in the entity.

This should at least allow the entire hash to be accessed with string keys yet default to existing behavior (symbol keys).

If anyone has suggestions for how to improve this I'm all ears. This was just the quickest least intrusive thing I could think of at the time.


LeFnord:
thanks @mhenrixon

PR #329, opened 2020-01-21
Option expose_nil doesn't work when block is passed

serbiant:
Greetings,

I've been trying to implement a conditional exposure at one of the entities in my project.
Noticed that in the current implementation, this code 

```ruby
expose :personal_details, expose_nil: false do |address, options|
  if options[:one_option].present?
    Public::Entities::PersonalDetail.represent address.personal_details
  elsif options[another_option].present?
    Public::Entities::PersonalDetailShort.represent address.personal_details
  else
    nil # explicit for the example
  end
end
```
will eventually expose `null` if both conditions are not satisfied.

This PR fixes the specified issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#329](https://github.com/ruby-grape/grape-entity/pull/329): Option expose_nil doesn't work when block is passed - [@serbiant](https://github.com/serbiant).
 * [#320](https://github.com/ruby-grape/grape-entity/pull/320): Gemspec: drop eol'd property rubyforge_project - [@olleolleolle](https://github.com/olleolleolle).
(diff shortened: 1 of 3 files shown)


LeFnord:
thanks @serbiant

Morxander:
@serbiant I'm using the latest version `0.8.0` which includes your fix but it's still exposing `null`.

NoryDev:
I think you broke something that was working.

Before this change:

```ruby
class PostEntity < Grape::Entity
  expose :title
end

class CommentEntity < Grape::Entity
  expose :post, expose_nil: false do |object, options|
    PostEntity.represent(object.post, options)
  end
end
```
Now let's imagine a `comment` with no `post` for some reason, in that scenario, `PostEntity.represent(object.post, options)` is not `nil`, it's an instance of `PostEntity`. But it evaluates to `nil` as `object.post` is nil

This used to work fine before this change, but now it doesn't as it thinks the block does not return nil. I think you just check nil evaluation in a different place.

PR #318, opened 2019-04-01
CI: Use 2.4.6, 2.5.5, 2.6.3

olleolleolle:
This PR updates the CI matrix.

PR #330, opened 2020-02-17
CI: Use 2.5.7,  2.6.5, 2.7.0

budnik:
Previous attempt to [update matrix](#318) failed because rubocop version is not fixed so `~> 0.65` resolved to version above `0.75`.  I guess it was mistake not to define it as `~> 0.65.0` 

Last version of rubocop passing is `0.74` so fixing that version.
Bumping of rubocop could be addressed in own PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ rvm:
   - 2.4.9
   - 2.5.7
   - 2.6.5
+  - 2.7.0
   - ruby-head
   - jruby-head
 
@@ -17,5 +18,6 @@ matrix:
   fast_finish: true
 
   allow_failures:
+    - rvm: 2.4.9
     - rvm: ruby-head
     - rvm: jruby-head
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ after_success:
   - bundle exec danger
 
 rvm:
-  - 2.4.5
-  - 2.5.4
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -7,9 +7,9 @@ after_success:
   - bundle exec danger
 
 rvm:
-  - 2.4.5
-  - 2.5.4
-  - 2.6.2
+  - 2.4.9
```
LeFnord: please move it it to allowd failures and add 2.7.0

Gemfile
```diff
@@ -5,7 +5,7 @@ source 'http://rubygems.org'
 gemspec
 
 group :development, :test do
-  gem 'rubocop', '~> 0.65', require: false
+  gem 'rubocop', '~> 0.74.0', require: false
```
LeFnord: mmh, can 0.79.0 be used?

LeFnord:
will close #318 in favour of this one

olleolleolle:
I really appreciate the work on bringing this forward! ✨

Gemfile
budnik: Yes but it needs to change few lines of code. I'll address it in separate PR if you don't mind. WDYT?

LeFnord: seems reasonable

LeFnord:
thanks @budnik

PR #331, opened 2020-02-17
Adds missing dash

budnik:
On every new PR Danger changelog check is complaining because of missing dash:

Screenshot 2020-02-17 at 14 03 38

---

I'm not sure if line in question should be both in fixes and features. But it got to have dash before author's name.


LeFnord:
thanks @budnik for fixing it

PR #333, opened 2020-02-18
Fix typo in CHANGELOG.md

eitoball:


LeFnord:
thanks @eitoball … the entry itself isn't really nedded … but yeah

PR #338, opened 2020-09-04
Fix ruby 2.7 deprecation warning

begotten63:
Fix warning with keyword arguments:

`/Users/begotten63/.rvm/gems/ruby-2.7.1/gems/grape-entity-0.8.1/lib/grape_entity/entity.rb:540: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`

`/Users/begotten63/.rvm/gems/ruby-2.7.1/gems/grape-entity-0.8.1/lib/grape_entity/delegator/hash_object.rb:7: warning: The called method 'delegate' is defined here`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Fixes
 
-* Your contribution here.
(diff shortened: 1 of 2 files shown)


LeFnord:
thanks @begotten63

PR #350, opened 2021-03-21
Fix typo in comments in lib/grape_entity/entity.rb

kfukai23:
I found a comment that seems to be a typo, so I suggest a fix.


pboling (approved):
LGTM

LeFnord:
thanks

Issue #327, opened 2019-12-04
Adding default value to expose 


rahul404:
Is there a way to add a default value to `expose` in case of the value is not present?
For example
```
expose :tags do |filter|
  filter[:tags].present? ? filter[:tags] : "all"
end
```

could be replaced by 
`expose :tags, default: "all"`

PR #352, opened 2021-06-11
Add default value option

ahmednaguib:
### Description
I noticed there is no direct way of defining default value for `expose` when a field is returning `nil`. Of course there is a way to do it via a block but would be a nice option to have it like this:
`expose :x, default: 0`

I also saw there is an issue opened https://github.com/ruby-grape/grape-entity/issues/327 asking for the feature so I made the pr maybe it will be useful for other folks.

### DIFF
--- a/lib/grape_entity/exposure/base.rb
+++ b/lib/grape_entity/exposure/base.rb
@@ -16,6 +16,7 @@ module Grape
           key = options[:as] || attribute
           @key = key.respond_to?(:to_sym) ? key.to_sym : key
           @is_safe = options[:safe]
+          @default_value = options[:default]
           @for_merge = options[:merge]
           @attr_path_proc = options[:attr_path]
           @documentation = options[:documentation]
@@ -82,7 +83,10 @@ module Grape
         end
 
         def valid_value(entity, options)
-          value(entity, options) if valid?(entity)
+          return unless valid?(entity)
+
+          output = value(entity, options)
+          output.blank? && @default_value.present? ? @default_value : output
         end
 
         def should_return_key?(options)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Your contribution here.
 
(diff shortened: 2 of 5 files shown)


lib/grape_entity/exposure/base.rb
```diff
@@ -82,7 +83,10 @@ def serializable_value(entity, options)
         end
 
         def valid_value(entity, options)
-          value(entity, options) if valid?(entity)
+          return unless valid?(entity)
+
+          output = value(entity, options)
+          output.blank? && @default_value.present? ? @default_value : output
```
dlinch, line 89: I am not a maintainer, but I stumbled across this PR while I was looking for a sane default ability for entities.
`blank?` and `present?` are Rails specific, so to keep Grape framework agnostic you might want to shuffle this around. In fact, the `blank?` check replaces some valid options like `false` and `[]` with the default value, which I would not expect.

```suggestion
output.nil? ? @default_value : output
```
I'm not sure the `.present?` check bought you anything anyway if we keep it strictly to a nil check.

LeFnord: `.blank?` and `.present?` are part of ActiveSupport and this is included in the Grap gemspec, so it can be used without adding additional dependencies

but suggest to move the getting of the value into the same named method → [`value()`](https://github.com/ruby-grape/grape-entity/blob/master/lib/grape_entity/exposure/base.rb#L64)

LeFnord:
sorry @ahmednaguib for the late reply

lib/grape_entity/exposure/base.rb
ahmednaguib: @LeFnord not sure if I get this part: 
> but suggest to move the getting of the value into the same named method → value()

dlinch: @LeFnord Do you still consider the behavior of `blank?` to be appropriate though? If I had a boolean attribute with a default of `true`, then the above implementation will always return `true` no matter what, as:
```ruby
false.blank?
# => true
```

It seems to me like the only sane value to override with a default would be `nil`

PR #355, opened 2021-08-25
Fix infinite loop problem with the `NameError`s in block exposures

meinac:
Calling `NameError#message` causes the `did_you_mean` library to endup
being in an infinite loop(Probabaly raises `SystemStackError` at some
point though).

Related to https://github.com/ruby/did_you_mean/issues/158


meinac:
Hi @LeFnord, do you have a plan to release a new version containing this fix?

LeFnord:
yeap, today, see https://github.com/ruby-grape/grape-entity/pull/356

PR #366, opened 2022-07-26
Don't suppress regular ArgumentError exceptions

splattael:
Account specifically for the following scenario:

```ruby
  expose :foo, &:bar
```

Note that `:bar.to_proc.parameters` always returns `[[:req], [:rest]]`.

We should not swallow ArgumentError exceptions in any other instance.

See spec for an example.


splattael:
:wave: @LeFnord Do you mind reviewing this PR? :pray:

LeFnord (approved):
👍 … that makes sense … good improvement

LeFnord:
@splattael … please, can you run `bundle exec rubocop --auto-gen-config`

splattael:
@LeFnord Whoops, done :heavy_check_mark: Thanks for the heads-up :bow:

LeFnord:
avoid calling of `SimpleCov` stuff

and you can also change [`ruby-version: ['2.6', '2.7', '3.0', head, jruby, truffleruby]`](https://github.com/ruby-grape/grape-entity/blob/master/.github/workflows/ruby.yml#L16) into `ruby-version: ['2.6', '2.7', '3.1', head, jruby, truffleruby]`

splattael:
@LeFnord 

> avoid calling of SimpleCov stuff

What do you mean by this? :thinking: 

> and you can also change [ruby-version: ['2.6', '2.7', '3.0', head, jruby, truffleruby]](https://github.com/ruby-grape/grape-entity/blob/master/.github/workflows/ruby.yml#L16) into ruby-version: ['2.6', '2.7', '3.1', head, jruby, truffleruby]

So, should we replace `3.1` with `3.0`? Why should we remove `3.0` or should we only add `3.1`? :thinking:

LeFnord:
> @LeFnord
> 
> > avoid calling of SimpleCov stuff
> 
> What do you mean by this? 🤔

sorry … mean it here [spec_helper.rb](https://github.com/ruby-grape/grape-entity/blob/master/spec/spec_helper.rb#L12)

> 
> > and you can also change [ruby-version: ['2.6', '2.7', '3.0', head, jruby, truffleruby]](https://github.com/ruby-grape/grape-entity/blob/master/.github/workflows/ruby.yml#L16) into ruby-version: ['2.6', '2.7', '3.1', head, jruby, truffleruby]
> 
> So, should we replace `3.1` with `3.0`? Why should we remove `3.0` or should we only add `3.1`? 🤔

replace

splattael:
> sorry … mean it here [spec_helper.rb](https://github.com/ruby-grape/grape-entity/blob/master/spec/spec_helper.rb#L12)

I still don't follow. Should I delete SimpleCov? :thinking: 

> replace

OK, done but honestly I don't understand why we stop testing on Ruby 3.0 :shrug: It is still being [maintained](https://www.ruby-lang.org/en/downloads/branches/).

LeFnord:
sorry my fault … yes you are right 3.0 is maintained … so re-add 3.0 

and comment out the `SimpleCov` stuff here [spec_helper.rb](https://github.com/ruby-grape/grape-entity/blob/master/spec/spec_helper.rb#L12) so the checks will be green and I can mörge it

splattael:
> sorry my fault … yes you are right 3.0 is maintained … so re-add 3.0

No worries, done :+1: 

> and comment out the SimpleCov stuff here [spec_helper.rb](https://github.com/ruby-grape/grape-entity/blob/master/spec/spec_helper.rb#L12) so the checks will be green and I can mörge it

Oh now I saw it fails on Ruby >= 3.1 due to https://github.com/simplecov-ruby/simplecov/issues/1003 and thus cancels all other runs :facepalm: 

I've disabled SimpleCov on Ruby >= 3.1 then so checks should pass soon :crossed_fingers:

Issue #368, opened 2022-09-03
`FetchableObject` can cause unexpected behavior


danielvdao:
### Context

When delegating to a `FetchableObject`, we've seen that it can cause breaking changes since it's only checking `object.respond_to?(:fetch, true)` [here](https://github.com/ruby-grape/grape-entity/blob/6eac458ddd69117ad979aaab7d1798bb1ba5a349/lib/grape_entity/delegator.rb#L18) as opposed to checking if the object is a `Hash`.

The reason this came up is because we depend on another gem [StoreModel](https://github.com/DmitryTsepelev/store_model) which delegates `fetch` to the underlying `Hash`. However the class wraps over a `Hash` with `ActiveModel`, so the `fetch` delegation isn't 100%  representative of calling `object.random_method`. 

This can cause a bug with `KeyError` as the underlying object is a `Hash`, but doesn't actually have the key on it. Instead we depend on the accessor method which does its own nil-safe handling behavior.

I'm looking into it here: https://github.com/DmitryTsepelev/store_model/pull/128, and I was curious about the inspiration for `FetchableObject` (since I may be missing something 😓) and whether or not it makes sense to actually remove it as I _feel_ like `fetch` should **only** be called on a `Hash`. In the PR linked it can get tricky to re-implement `#fetch` properly 😓 

I have an example commit https://github.com/danielvdao/grape-entity/commit/92df52be26f26ee7f916c5495130d841f51fe77e, and can open a PR / actually remove `FetchableObject` if that's the right approach, but I wanted to get y'all's thoughts first.

danielvdao:
Hi @LeFnord I was wondering if you happened to have context on this? I'm happy to remove `FetchableObject`, but I'm curious if there's something about it that I may be missing.

LeFnord:
mmh, you can try it

but have in mind, this will be used by expose, to find out how the data can be accessed, means `object.fetch` should be available

danielvdao:
@LeFnord I'll open a PR. I think the confusion is that some objects respond to `fetch`, but don't match with their `public_send` behavior. I would imagine that `#fetch` should only be used for `Hash` 🤔 but in that case we already have `Hash`.

danielvdao:
@LeFnord let me know if this makes sense? Thank you!

https://github.com/ruby-grape/grape-entity/pull/369

PR #369, opened 2022-09-15
🔥 Remove `FetchableObject`

danielvdao:
### Context

A PR addressing this issue: https://github.com/ruby-grape/grape-entity/issues/368

The thing I am confused about is why we check if `object.respond_to?(:fetch, true)`. Implementing `#fetch` is tricky and I think that most of the time (if not always) we should rely on `Hash` to implement `#fetch` as that's in Ruby stdlib. Otherwise just default to `#public_send`.

Happy to close if it doesn't make sense either - but wanted to give it a shot and it looks like all tests still pass.

cc @LeFnord


danielvdao:
Thanks, I'll take care of it 🔥

danielvdao:
@LeFnord done! Let me know if this is what you were looking for.

danielvdao:
Hey @LeFnord some of the Rubocop violations are going off, but I don't think they're related to my PR. What do you think? 🤔

LeFnord:
mmh, will have an eye on

spec/grape_entity/entity_spec.rb
```diff
@@ -29,9 +29,11 @@
         end
 
         it 'makes sure that :format_with as a proc cannot be used with a block' do
-          # rubocop:disable Style/BlockDelimiters
-          expect { subject.expose :name, format_with: proc {} do p 'hi' end }.to raise_error ArgumentError
-          # rubocop:enable Style/BlockDelimiters
+          expect do
```
danielvdao, line 32: I cleaned this one up, but I wasn't sure about the other development dependency ones.

LeFnord:
I'm on it

LeFnord:
thanks again

PR #371, opened 2022-11-29
Allow default exposed value to be `false` or any empty data

norydev:
Using `present?` limits what can be set as `:default` and restricts falsy values like `false`, `''`, `[]` for example.

I don't really see why not allow those (I have a use case for `false` for example, which is to ensure a nullable boolean value is always rendered as a boolean).


pawelma (approved):

LeFnord:
ok makes sense … thanks @norydev

PR #376, opened 2023-07-27
Fix the method Grape::Entity#exec_with_object to work with Ruby 3

linhchauatx:
Fix the method Grape::Entity#exec_with_object to work with Ruby 3.

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -514,7 +514,7 @@ module Grape
     end
 
     def exec_with_object(options, &block)
-      if block.parameters.count == 1
+      if block.parameters.count == 1 || block.parameters == [[:req], [:rest]]
         instance_exec(object, &block)
       else
         instance_exec(object, options, &block)
@@ -523,6 +523,7 @@ module Grape
       # it handles: https://github.com/ruby/ruby/blob/v3_0_0_preview1/NEWS.md#language-changes point 3, Proc
       # accounting for expose :foo, &:bar
       if e.is_a?(ArgumentError) && block.parameters == [[:req], [:rest]]
+        Rails.logger.error("***** ERROR in exec_with_object: #{e.message}\n#{e.backtrace.join("\n")}")
         raise Grape::Entity::Deprecated.new e.message, 'in ruby 3.0'
       end
 


dblock:
We already have Ruby 3.x in the CI. Are we missing tests? (Add part of this PR please)?

Don't worry about versions, revert that.

lib/grape_entity/entity.rb
```diff
@@ -523,6 +523,7 @@ def exec_with_object(options, &block)
       # it handles: https://github.com/ruby/ruby/blob/v3_0_0_preview1/NEWS.md#language-changes point 3, Proc
       # accounting for expose :foo, &:bar
       if e.is_a?(ArgumentError) && block.parameters == [[:req], [:rest]]
+        Rails.logger.error("***** ERROR in exec_with_object: #{e.message}\n#{e.backtrace.join("\n")}")
```
dblock, line 526: Not everyone uses Rails.

dblock (changes requested):
This has a fix but also still has a `Rails` mention, needs work.

numbata:
After working through the #389  and seeing how our `&:method` shorthand behaves, I've come to believe that maybe it's better to deprecate this feature entirely. Here's why:

Right now:
```ruby
class FooObject
  def method_without_args
    'result'
  end
end

class FooObjectEntity < Grape::Entity
  # These two lines do exactly the same thing:
  expose :boo,                &:method_without_args
  expose :method_without_args, as: :boo_again
end

foo = FooObject.new
FooObjectEntity.represent(foo).serializable_hash
# => { boo: "result", boo_again: "result" }
```
In other words:
```ruby
expose :alias, &:method_name
# is identical to:
expose :method_name, as: :alias
```

Because there’s no extra behavior or transformation happening, the shorthand doesn't buy us anything over the already-concise as: form. If our goal is in "expose a zero-argument method under a different key," we can do that directly.

The usage of `&:method` causes confusion and extra complexity in context of grape-entity:

1. Misleading “format” vs. “attribute” semantics
Some users could expect `&:method` to behave like a formatter (e.g. “call my own entity helper named method”), but in fact it always calls the method on the model. That confusion led to a lot of special casing in `exec_with_object`)
2. Always requires zero arguments
We now have to detect “symbol-to-proc” blocks, lookup the original method’s arity, and raise if it isn’t zero. That logic lives in `exec_with_object` and complicates our code path.
3. No real shorthand advantage
Compared to `expose :foo, as: :bar`, the only thing `&:method_name` does is skip the `as:` keyword. But you still end up writing two tokens(`:foo, &:bar`) instead of just more transparent `:bar, as: :foo`.

Because of these points, the "nice little shorthand" is actually increasing our maintenance burden and confusing people more than helping them (in my subjective opinion).

Another idea is to make `&:method` behave more like a mini-formatter—something akin to `format_with:`. For instance:

```ruby
class FooObject
  def name    = "name"
  def surname = "surname"
end

class FooObjectEntity < Grape::Entity
  # Treat `&:capitalize` as “call my own `capitalize` helper”
  expose :name, &:capitalize
  expose :surname, &:capitalize

  # If `value` is the raw attribute value (e.g. "name"),
  # `object` and `options` can be used, too.
  def capitalize(value, object = nil, options = nil)
    value.upcase
  end
end

foo = FooObject.new
FooObjectEntity.represent(foo).serializable_hash
# => { name: "NAME", surname: "SURNAME" }
```

In this proposal, `&:capitalize` doesn’t mean "call object.capitalize" (which might not even exist), but instead delegates to a method defined inside the entity class. That would give us a lightweight way to define per-entity formatting logic without needing a full `format_with` block.

wdyt?

PR #381, opened 2023-12-13
Pass the right options when determining whether to expose an attribute using a block

### LINKED ISSUE #378: `expose_nil:` conflicts with options passing to a block
Minimum reproducible example:
```ruby
class Test < Grape::Entity
  expose(:a, expose_nil: false) { |obj, options| options[:a] }
end

pp Test.new({}, a: 1).as_json # expected {a: 1}, renders {}
pp Test.new({}, a: nil).as_json # expected {}, renders {}
```
More debugging shows that block always returns `nil`, because when `expose_nil:` is present, the content of `options` inside block is changed completely; which this debugging serializer shows:
```ruby
class Test < Grape::Entity
   expose(:a) { |obj, options| p(simple: options); options[:a] }
   expose(:b, as: :d) { |obj, options| p(with_as: options); obj[:a] }

   expose(:c, expose_nil: false) { |obj, options| p(with_expose_nil: options); options[:a] }
end

pp Test.new({}, a: 1, b: 2, c: 3).as_json
```
It prints this values for options inside blocks (and yes, in this order):
```
{:with_expose_nil=>{:expose_nil=>false, :proc=>#<Proc:0x0000555bb661d9a8 test_grape_expose.rb:52>}}
{:simple=>#<Grape::Entity::Options:0x0000555bb661c210 @opts_hash={:a=>1, :b=>2, :c=>3, :attr_path=>[:a]}, @has_only=false, @has_except=false, @for_nesting_cache={}, @should_return_key_cache={}>}
{:with_as=>#<Grape::Entity::Options:0x0000555bb661c210 @opts_hash={:a=>1, :b=>2, :c=>3, :attr_path=>[:d]}, @has_only=false, @has_except=false, @for_nesting_cache={}, @should_return_key_cache={}>}
```
It seems that block in this case receives "options passed to `expose` DSL" instead of "options passed to serialization"?..

magni-:
As @zverok deduced when [opening the issue](https://github.com/ruby-grape/grape-entity/issues/378#issue-1853710374), the wrong options are getting used when pairing `expose_nil: false` with a block.

Fixes #378

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ AllCops:
     - vendor/**/*
     - example/**/*
   NewCops: enable
-  TargetRubyVersion: 3.2
+  TargetRubyVersion: 3.0
   SuggestExtensions: false
 
 #  Layout stuff
--- a/lib/grape_entity/exposure.rb
+++ b/lib/grape_entity/exposure.rb
@@ -54,7 +54,7 @@ module Grape
 
         def expose_nil_condition(attribute, options)
           Condition.new_unless(
-            proc do |object, _options|
+            proc do |object, entity_options|
               if options[:proc].nil?
                 delegator = Delegator.new(object)
                 if is_a?(Grape::Entity) && delegator.accepts_options?
@@ -63,7 +63,7 @@ module Grape
                   delegator.delegate(attribute).nil?
                 end
               else
-                exec_with_object(options, &options[:proc]).nil?
+                exec_with_object(entity_options, &options[:proc]).nil?
               end
             end
           )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#381](https://github.com/ruby-grape/grape-entity/pull/381): Fix `expose_nil: false` when using a block - [@magni-](https://github.com/magni-).
 * Your contribution here.
 
 
--- a/lib/grape_entity/condition/base.rb
+++ b/lib/grape_entity/condition/base.rb
@@ -4,8 +4,8 @@ module Grape
   class Entity
     module Condition
       class Base
-        def self.new(inverse, *args, &block)
-          super(inverse).tap { |e| e.setup(*args, &block) }
+        def self.new(inverse, ...)
+          super(inverse).tap { |e| e.setup(...) }
         end
 
         def initialize(inverse = false)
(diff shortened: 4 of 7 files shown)


lib/grape_entity/exposure.rb
```diff
@@ -63,7 +63,7 @@ def expose_nil_condition(attribute, options)
                   delegator.delegate(attribute).nil?
                 end
               else
-                exec_with_object(options, &options[:proc]).nil?
+                exec_with_object(entity_options, &options[:proc]).nil?
```
LeFnord, line 66: good spot

.rubocop.yml
```diff
@@ -74,6 +74,10 @@ Naming:
 
 #  Style stuff
 #
+Style/ArgumentsForwarding:
+  # disabled because this breaks on Ruby < 3.2
```
LeFnord, line 78: mmh, I'm not sure if this is really needed, cause expecting ruby > 3.0 supports it

LeFnord (approved):
thanks @magni- … nice catch

.rubocop.yml
magni-: See https://github.com/ruby-grape/grape-entity/actions/runs/7191884680

Anonymous args/kwargs forwarding was added in Ruby 3.2.

zverok: BTW, it seems that a better solution is to change [this line](https://github.com/ruby-grape/grape-entity/blob/master/.rubocop.yml#L8) to 3.0.

It means to designate the _minimum_ Ruby version that should be supported by the code (so that Rubocop wouldn't propose features that emerged in later versions).

LeFnord: no not really, want to support the future not the past

.rubocop.yml
```diff
@@ -5,7 +5,7 @@ AllCops:
     - vendor/**/*
     - example/**/*
   NewCops: enable
-  TargetRubyVersion: 3.2
+  TargetRubyVersion: 3.0
```
LeFnord, line 8: please revert it

.rubocop.yml
zverok: @LeFnord The [exact meaning](https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version) of this field is to set the _minimum_ version that the codebase is intended to support. If you have a codebase that needs to support 3.0 and 3.1, and you set this field to 3.2, then Rubocop will suggest to you various Ruby features that _only_ work on 3.2 and make the code broken on 3.1 (the exact problem that happened in this PR).

LeFnord: for compatibility we have a test matrix, rubocop is used for actual an future dev … but yeah, it seems, it makes no sense to discus

mscrivo:
@LeFnord could I trouble you to put out a new release of grape-entity? We've run into this issue. Alternatively, I'd be happy to do the legwork of putting out a release if you or @dblock can add me as a maintainer.

LeFnord:
done

PR #384, opened 2024-04-28
Update inspect method to handle nil serializable_hash

fcce:
`serializable_hash` may return `nil` 

before 
```ruby
Entities::User.represent nil
# An error occurred when inspecting the object: #<NoMethodError: undefined method `map' for nil:NilClass>
```
after 

```ruby
Entities::User.represent nil
#<Entities::User:326200 nil>
```

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -472,8 +472,13 @@ module Grape
 
     # Prevent default serialization of :options or :delegator.
     def inspect
-      fields = serializable_hash.map { |k, v| "#{k}=#{v}" }
-      "#<#{self.class.name}:#{object_id} #{fields.join(' ')}>"
+      hash = serializable_hash
+      if hash.nil?
+        "#<#{self.class.name}:#{object_id}> nil"
+      else
+        fields = object.map { |k, v| "#{k}=#{v}" }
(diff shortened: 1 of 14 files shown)


dblock (changes requested):
Add a test and update CHANGELOG please?

dblock (approved):

dblock:
@numbata or @fcce want to finish this one?

lib/grape_entity/entity.rb
```diff
@@ -472,8 +472,13 @@ def presented
 
     # Prevent default serialization of :options or :delegator.
     def inspect
-      fields = serializable_hash.map { |k, v| "#{k}=#{v}" }
-      "#<#{self.class.name}:#{object_id} #{fields.join(' ')}>"
+      object = serializable_hash
+      if object.nil?
+        "#<#{self.class.name}:#{object_id}> nil"
```
numbata, line 477: As I know, the value of object preferable to keep inside the `<>`, so maybe this line shouldn't be like:
```suggestion
        "#<#{self.class.name}:#{object_id} nil>"
```
wdyt? 🤔

dblock:
Merged this, thanks for your help @numbata!

PR #385, opened 2024-05-05
Drop `multijson` dependency

ericproulx:
This PR adds a `Grape::Entity::Json` class that will use `multijson` if already loaded instead of the default `json`.

A spec has been added to make sure `to_json` still works and moved `require 'spec_helper'` in `.rspec` file

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,4 @@
+--require spec_helper
 --color
 --profile
 --format documentation
--- a/lib/grape_entity/json.rb
+++ b/lib/grape_entity/json.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Grape
+  class Entity
+    if defined?(::MultiJson)
+      Json = ::MultiJson
+    else
+      Json = ::JSON
+      Json::ParseError = Json::ParserError
+    end
+  end
+end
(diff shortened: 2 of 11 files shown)


ericproulx:
@LeFnord If you want to take a look :)

lib/grape_entity/json.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Grape
+  class Entity
+    if defined?(::MultiJson)
+      Json = ::MultiJson
+    else
+      Json = ::JSON
+      Json::ParseError = Json::ParserError
```
LeFnord, line 9: why this line?

.rspec
```diff
@@ -1,3 +1,4 @@
+--require spec_helper
```
LeFnord, line 1: it isn't really related, but nice … 👍

LeFnord (commented):
thanks @ericproulx … only one question

lib/grape_entity/json.rb
ericproulx: I took the code from grape directly. In fact, since grape-entity is not parsing, we could drop it.

LeFnord: ok

LeFnord (approved):
thanks @ericproulx

PR #388, opened 2025-05-22
Drop ruby-head from test matrix to keep builds stable

numbata:
This trims the workflow by dropping the nightly ruby-head jobs. The snapshot build is flaky right now, so it hides real regressions and slows CI without helping users. Until Ruby 3.5 is out (or close to that point), we’ll run only the stable 3.1-3.4 releases; 3.5 will be added in a separate PR. No runtime code change - just faster, greener builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     needs: ['rubocop']
     strategy:
       matrix:
-        ruby-version: ['3.0', '3.1', '3.2', '3.3', 'head']
+        ruby-version: ['3.0', '3.1', '3.2', '3.3', '3.4']
     steps:
     - name: Check out branch
       uses: actions/checkout@v4


.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     needs: ['rubocop']
     strategy:
       matrix:
-        ruby-version: ['3.0', '3.1', '3.2', '3.3', 'head']
+        ruby-version: ['3.0', '3.1', '3.2', '3.3', '3.4']
```
LeFnord, line 31: remove 3.0, as you mentioned in the description, and maybe replacing head by 3.5-dev

numbata: We currently have 3.0 as a requirement in the gemspec.
What if I just change the PR description for now. And add 3.5-dev with ruby-head in a separate pull request where we can focus on supporting exactly 3.5?

dblock:
Merging this, thanks for fixing CI!

PR #389, opened 2025-06-01
Block arity detection

numbata:
Ruby 3’s stricter Proc arity rules could slip unwanted arguments into &:method_name exposures, causing confusing errors. This change detects pure symbol-to-proc wrappers up front, looks up the actual method’s arity on the entity object, and raises an error if it’s not zero. Regular blocks still use their own arity to decide whether to pass options. In practice, this makes it impossible to accidentally expose a one- or two-argument method with &:foo, ensuring only truly zero-argument methods are used in that shorthand.

Possible sollution #376

### DIFF
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -518,26 +518,41 @@ module Grape
       root_exposure.serializable_value(self, opts)
     end
 
-    def determine_block_arity(block)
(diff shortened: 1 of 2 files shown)


numbata:
More details in https://github.com/ruby-grape/grape-entity/pull/376#issuecomment-2932021080

PR #398, opened 2026-04-15
Update CHANGELOG with missing entries for version 1.0.2

faheemmughal:
Added entries for missing PRs within the version. Especially for a breaking change that is failing 400+ tests for us.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,12 @@
 * [#394](https://github.com/ruby-grape/grape-entity/pull/394): Add `Entity.[]` for Grape >= 3.2 param type compatibility - [@numbata](https://github.com/numbata).
 * [#388](https://github.com/ruby-grape/grape-entity/pull/388): Drop ruby-head from test matrix - [@numbata](https://github.com/numbata).
 * [#384](https://github.com/ruby-grape/grape-entity/pull/384): Fix `inspect` to correctly handle `nil` values - [@fcce](https://github.com/fcce).
+* [#385](https://github.com/ruby-grape/grape-entity/pull/385): Drop multijson dependency - [ericproulx](https://github.com/ericproulx).
+
+
+**Breaking change:**
+* [#389](https://github.com/ruby-grape/grape-entity/pull/389): Block arity detection - [@numbata](https://github.com/numbata).
+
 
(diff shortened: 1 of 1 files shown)


numbata:
@faheemmughal Thanks for flagging this, really appreciate it.
I’ll add a Danger check to make sure we don’t miss CHANGELOG updates going forward.

Could you share a bit more detail about the “400+ failing tests” you mentioned? Specifically:
  * which versions of `grape` and `grape-entity` you’re using
  * whether you have any related gems (e.g. `grape-swagger`) in the setup

That should help me understand what’s going on 👍

faheemmughal:
Thank you for being so responsive. 

In our codebase we are using variations of the following, 

```ruby 
expose :property, if: ->(model) { model.active? }

expose :storage,
       if: ->(model) { model.storage_method.present? } do |model|
  check_storage(model, :storage_method)
end
```

Update to version 1.0.2 is causing test failures for us: 

```bash
  ArgumentError:
       wrong number of arguments (given 2, expected 1)
```

And forcing us to change the code to,

```ruby
expose :property, if: ->(model, _) { model.active? }
expose :storage,
       if: ->(model, _) { model.storage_method.present? } do |model, _|
  check_storage(model, :storage_method)
end
```

I haven't looked too closely at it but it's likely that the change in https://github.com/ruby-grape/grape-entity/pull/389 is the cause.

numbata:
Thanks for the report and especially for the examples. Sorry for the confusion; this is definitely not the experience we want from the gem.

I’ll prepare a patch release to restore the previous behavior as soon as possible.

numbata:
@faheemmughal l The fix is up in #399. Both your original single-argument style and the two-argument workaround will keep working, so no need to roll back any changes you've already made.

Will release 1.0.3 soon.

numbata:
[v1.0.3 is out](https://rubygems.org/gems/grape-entity/versions/1.0.3) with the fix from #399. I've incorporated the `CHANGELOG` updates [separately](https://github.com/ruby-grape/grape-entity/pull/400/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed), so closing this PR. 

Thanks again for flagging the issue so quickly! And let us know if you run into anything else.

PR #404, opened 2026-04-17
Fix Time serialization regression from dropping MultiJson

### LINKED ISSUE #403: Dropping MultiJson in 1.0.3 broke Time serialization for apps using ActiveSupport
## Description

 After upgrading from grape-entity 1.0.1 to 1.0.3, all `Time` fields in API responses changed format for apps using ActiveSupport:

  - **Before (1.0.1):** `"2026-04-16T10:55:10.597Z"` (ISO 8601)
  - **After (1.0.3):** `"2026-04-16 10:55:10 UTC"`

  This is a silent breaking change for API consumers expecting ISO 8601 timestamps.

  ## Root cause

  PR #385 replaced `MultiJson.dump` with `JSON.dump` in `Entity#to_json`: https://github.com/ruby-grape/grape-entity/pull/385/changes#diff-dede7a9b2a44333c10da3b7b006d011d8fddfc23ebaa64b415b5686c7798d296R561

  `MultiJson.dump` internally calls `Hash#to_json`, which ActiveSupport intercepts to run `as_json` on all values — converting `Time` to ISO 8601. `JSON.dump` bypasses ActiveSupport entirely and serializes `Time` via its C extension using `Time#to_s`.

  ## Reproduction

  ```ruby
  require 'json'
  require 'active_support'
  require 'active_support/core_ext/object/json'
  require 'active_support/core_ext/time'
  require 'grape_entity'

  class TestEntity < Grape::Entity
    expose :published_at
  end

  t = Time.utc(2026, 4, 16, 10, 55, 10, 597000)
  obj = OpenStruct.new(published_at: t)

  TestEntity.represent(obj).to_json
  # grape-entity 1.0.1  => {"published_at":"2026-04-16T10:55:10.597Z"}  ✅
  # grape-entity 1.0.3 => {"published_at":"2026-04-16 10:55:10 UTC"}   ❌
  ```

  ## Environment

  - Ruby 4.0.0
  - Rails 8.0.5
  - grape-entity 1.0.3

numbata:
## Summary

Version 1.0.3 introduced a silent breaking change for all API consumers: `Time` fields switched from ISO 8601 format (`"2026-04-16T10:55:10.597Z"`) to Ruby's default `Time#to_s` format (`"2026-04-16 10:55:10 UTC"`). This breaks any client parsing timestamps from the API.

## What happened

PR #385 replaced `MultiJson.dump` with `JSON.dump` in `Entity#to_json` to drop the `multi_json` dependency. While the intent was correct, `JSON.dump` and `MultiJson.dump` behave differently when ActiveSupport is present:

- **`MultiJson.dump(hash)`** internally calls `Hash#to_json`, which ActiveSupport overrides to run `as_json` recursively on all values — giving `Time` the ISO 8601 representation.
- **`JSON.dump(hash)`** goes directly to Ruby's C extension JSON generator, bypassing ActiveSupport's `Hash#to_json` override entirely — so `Time` falls back to `Time#to_s`.

Since ActiveSupport is a runtime dependency of grape-entity, every user is affected.

An additional problem noted by @stevenou: the intermediate `Grape::Entity::Json` constant used `defined?(::MultiJson)` at require time to decide which backend to use. For apps with `gem "multi_json", require: false`, MultiJson wasn't loaded yet at that point, so it always fell back to `::JSON` regardless — making the MultiJson detection dead code in practice.

## How this fix works

Replaces `Grape::Entity::Json.dump(serializable_hash(options))` with `serializable_hash(options).to_json`.

`Hash#to_json` is the method ActiveSupport overrides. This single change resolves all reported problems:

1. **Time serialization** — `Hash#to_json` goes through ActiveSupport's `as_json` chain, restoring ISO 8601 output for `Time` and any other type with an `as_json` override.
2. **Nested exposures** — `OutputBuilder` (a `SimpleDelegator` wrapping a Hash) properly delegates `.to_json` to the underlying Hash.
3. **MultiJson load-order race** — eliminated entirely. We no longer check for or use MultiJson at all, so `require: false` scenarios are irrelevant.

The now-unused `Grape::Entity::Json` constant and `grape_entity/json.rb` are removed.

Fixes #403

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#404](https://github.com/ruby-grape/grape-entity/pull/404): Drop `MultiJson` dependency, use `Hash#to_json` for ActiveSupport-aware serialization - [@numbata](https://github.com/numbata).
 
 ### 1.0.4 (2026-04-17)
 
--- a/grape-entity.gemspec
+++ b/grape-entity.gemspec
@@ -25,7 +25,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 3.0'
 
   s.add_dependency 'activesupport', '>= 3.0.0'
-  s.add_dependency 'multi_json', '>= 1.0'
 
   s.files         = Dir['lib/**/*.rb', 'CHANGELOG.md', 'LICENSE', 'README.md']
   s.require_paths = ['lib']
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'multi_json'
-
 module Grape
   # An Entity is a lightweight structure that allows you to easily
   # represent data from your application in a consistent and abstracted
@@ -594,7 +592,13 @@ module Grape
 
     def to_json(options = {})
       options = options.to_h if options&.respond_to?(:to_h)
+<<<<<<< HEAD
       MultiJson.dump(serializable_hash(options))
+||||||| parent of 374a021 (Fix Time serialization regression introduced by dropping MultiJson)
+      Grape::Entity::Json.dump(serializable_hash(options))
+=======
+      serializable_hash(options).to_json
+>>>>>>> 374a021 (Fix Time serialization regression introduced by dropping MultiJson)
     end
 
     def to_xml(options = {})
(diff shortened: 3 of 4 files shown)


dblock:
I see you added tests for this in https://github.com/ruby-grape/grape-entity/pull/405, but I also think you should add a test that compares the serialization with a well-known string - it's possible that both `as_json` and `to_json` get broken together. Otherwise LGTM.

```ruby
expect(parsed['details']['birthday']).to eq('actual value')
```

PR #406, opened 2026-05-08
Feature: Handle symbol-to-proc wrappers (`&:`) that refer to a method created using delegation or method missing

marcrohloff:
## Summary

Improve handling of symbol-to-proc exposure blocks (`&:method_name`) for delegated and dynamically dispatched methods.

Previously grape-entity eagerly validated method arity for `&:method_name` blocks. This worked for ordinary methods, but produced incorrect behavior for methods implemented via `delegate` or `method_missing`, where Ruby may expose ambiguous or negative arity values.

This PR makes arity validation more conservative and Ruby-compatible:

* continue raising curated errors for ordinary methods with required arguments
* correctly allow methods with only optional args/splats/kwargs
* gracefully handle delegated and `method_missing`-backed methods
* fall back to Ruby's native runtime errors when introspection is unreliable

## Changes

* tighten symbol-to-proc wrapper detection
* add `positional_arity_for` helper for normalized arity handling
* support optional positional/keyword arguments correctly
* avoid false positives for delegation wrappers and dynamic dispatch
* add extensive spec coverage for delegation, `method_missing`, splats, kwargs, and edge cases

## Backward Compatibility

Public API is unchanged.

The only observable behavior change is that delegated or dynamic-dispatch methods may now raise Ruby-native errors instead of grape-entity-generated arity errors when the callable contract cannot be reliably introspected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### Next Release
 
-* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing`
+* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing` - [@marcrohloff](https://github.com/marcrohloff).
 
 #### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Next Release
 
+* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing`
+
 #### Features
 
 * Your contribution here.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 #### Features
 
-* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing` - [@marcrohloff](https://github.com/marcrohloff).
 * Your contribution here.
 
 #### Fixes
 
 * Your contribution here.
 * [#404](https://github.com/ruby-grape/grape-entity/pull/404): Drop `MultiJson` dependency, use `Hash#to_json` for ActiveSupport-aware serialization - [@numbata](https://github.com/numbata).
+* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing` - [@marcrohloff](https://github.com/marcrohloff).
 
 ### 1.0.4 (2026-04-17)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,8 @@
 ### Next Release
 
-* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing` - [@marcrohloff](https://github.com/marcrohloff).
-
 #### Features
 
+* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing` - [@marcrohloff](https://github.com/marcrohloff).
 * Your contribution here.
 
 #### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 * Your contribution here.
 * [#404](https://github.com/ruby-grape/grape-entity/pull/404): Drop `MultiJson` dependency, use `Hash#to_json` for ActiveSupport-aware serialization - [@numbata](https://github.com/numbata).
-* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing` - [@marcrohloff](https://github.com/marcrohloff).
+* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing`, and let unknown methods raise a native `NoMethodError` - [@marcrohloff](https://github.com/marcrohloff).
 
 ### 1.0.4 (2026-04-17)
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ Implement your feature or bug fix.
 
 Ruby style is enforced with [Rubocop](https://github.com/bbatsov/rubocop), run `bundle exec rubocop` and fix any style issues highlighted.
 
-Make sure that `bundle exec rake` and `rubocop` completes without errors.
+Make sure that `bundle exec rake` completes without errors.
 
 #### Write Documentation
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ Implement your feature or bug fix.
 
 Ruby style is enforced with [Rubocop](https://github.com/bbatsov/rubocop), run `bundle exec rubocop` and fix any style issues highlighted.
 
-Make sure that `bundle exec rake` completes without errors.
+Make sure that `bundle exec rake` and `rubocop` completes without errors.
 
 #### Write Documentation
 
--- a/lib/grape_entity/entity.rb
+++ b/lib/grape_entity/entity.rb
@@ -539,30 +539,46 @@ module Grape
       return unless match # Unrecognized format -> bail safe rather than misidentify
 
       origin_method_name = match[:name].to_sym
-      required_positional_arg_count, variadic = positional_arity_for(origin_method_name)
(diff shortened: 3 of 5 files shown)


lib/grape_entity/entity.rb
```diff
@@ -548,7 +548,8 @@ def ensure_block_arity!(block)
       end
 
       arity = object.method(origin_method_name).arity
-      return if arity.zero?
+      # functions defined using `delegate` or `method_missing` have an arity of -1
+      return if arity <= 0
```
numbata: This is too permissive and reintroduces the bug class #389 was meant to prevent. Ruby's negative arities encode `-(required + 1)`, so `def foo(a, *rest)` has `arity -2` and `def foo(a, b, *rest)` has arity `-3` - both have required positional
args, but both now sail through this guard and blow up later at `instance_exec(object, &block)` with a "wrong number of arguments" raised from inside user code, instead of the friendly `ArgumentError` this method was designed to produce.

Worth noting: `def method_missing(method, ...)` itself has arity `-2`, so even the stated motivating case is partially mishandled by `<= 0`.

Suggested fix is match the actual intent ("no required positional args"):

```ruby
arity = object.method(origin_method_name).arity
required = arity >= 0 ? arity : -arity - 1
return if required.zero?
```

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ### Next Release
 
+* [#406](https://github.com/ruby-grape/grape-entity/pull/406): Handle symbol-to-proc wrappers (`&:method_name`) where the method uses `delegate` or `method_missing` - [@marcrohloff](https://github.com/marcrohloff).
```
numbata: The entry is placed above the `#### Features` / `#### Fixes` headers rather than under one of them. Please move the line under `#### Fixes`.

CONTRIBUTING.md
```diff
@@ -44,7 +44,7 @@ Implement your feature or bug fix.
 
 Ruby style is enforced with [Rubocop](https://github.com/bbatsov/rubocop), run `bundle exec rubocop` and fix any style issues highlighted.
 
-Make sure that `bundle exec rake` completes without errors.
+Make sure that `bundle exec rake` and `rubocop` completes without errors.
```
numbata:
```suggestion
Make sure that `bundle exec rake` and `rubocop` complete without errors.
```

Subject-verb agreement: two subjects -> complete. Also the line above already instructs users to run bundle exec rubocop, so this becomes a near-duplicate 🤷

spec/grape_entity/entity_spec.rb
```diff
@@ -459,6 +484,28 @@ def raises_argument_error
             end
           end
 
+          context 'with block passed in via & that uses `missing_method`' do
+            specify do
+              subject.expose :using_missing, &:method_using_missing
+
+              object = SomeObject.new
+              expect(object.method(:method_using_missing).arity).to eq(-1)
```
numbata: The `value` assertion is the one that proves `grape-entity` handles it correctly. The arity line is testing Ruby.

```suggestion
```

spec/grape_entity/entity_spec.rb
```diff
@@ -415,6 +423,23 @@ def method_with_multiple_args(_object, _options)
             def raises_argument_error
               raise ArgumentError, 'something different'
             end
+
+            def method_missing(method, ...)
```
numbata, line 455: Rather than adding `method_missing` directly to `SomeObject`, could you extract the new fixtures into subclasses? That way the shared class stays clean and the existing tests (especially the "undefined method" one) don't depend on your
respond_to_missing? being correct to pass.

Something like:

```ruby
SomeObjectWithMissing = Class.new(SomeObject) do
  def method_missing(method, ...)
    return 'missing-result' if method == :method_using_missing
    super
  end

  def respond_to_missing?(method, include_private = false)
    method == :method_using_missing || super
  end
end

SomeObjectWithDelegation = Class.new(SomeObject) do
  InnerDelegate = Class.new { def delegated_method = 'delegated-result' }
  delegate :delegated_method, to: :inner
  def inner = @inner ||= InnerDelegate.new
end
```

and specs kind of:

```ruby
context 'with &: referencing a method_missing-backed method' do
  specify do
    subject.expose :using_missing, &:method_using_missing

    object = SomeObjectWithMissing.new
    expect(subject.represent(object).value_for(:using_missing)).to eq('missing-result')
  end
end

context 'with &: referencing a delegated method' do
  specify do
    subject.expose :using_delegation, &:delegated_method

    object = SomeObjectWithDelegation.new
    expect(subject.represent(object).value_for(:using_delegation)).to eq('delegated-result')
  end
end
```
with  the regression case that pins the boundary:
```ruby
context 'with &: referencing a method with required args and a splat (arity -2)' do
  specify do
    subject.expose :x, &:method_with_required_and_splat

    object = SomeObjectWithMissing.new  # or any object with such a method
    expect do
      subject.represent(object).value_for(:x)
    end.to raise_error(ArgumentError, match(/method expects/))
  end
end
```

numbata (changes requested):
@marcrohloff Thanks for the contribution! The motivation is solid. 👍 

Two things to address before merging: the `arity <= 0` guard is too wide (see inline comment), and the new `method_missing` in the shared fixture risks making the "undefined method" test fragile. Left suggestions on both.

spec/grape_entity/entity_spec.rb
marcrohloff: @numbata 

I decided to add some more specs to test the negative cases. Something like the others tests. i..e
```
subject.expose :that_method_with_required_and_optional_args, &:method_with_required_and_optional_args
subject.expose :method_with_required_and_optional_args, as: :that_method_with_required_and_optional_args_again
```

Getting the value of the first `expose` gives the expected exception message. But the second results in `wrong number of arguments (given 0, expected 1)` because it is processed through the delegator

Do you think that the second message should match the first message?

numbata: Good question! My short subjective answer is: no, they don't need to match for this PR.

The `ensure_block_arity!` only guards the `&:symbol` path — the `as:` alias goes through different machinery and its error message is a separate concern. Aligning the two would be a worthwhile follow-up, but let's not expand scope here.

For the context: `as:` bypasses `ensure_block_arity!` not by skipping it directly, but because it never produces a `proc` at all, so the exposure type is `DelegatorExposure` (plain attribute delegation) rather than `BlockExposure`. Two completely different code paths.

spec/grape_entity/entity_spec.rb
marcrohloff: It wasn't so much for testing as documenting an unexpected value. I removed it

CONTRIBUTING.md
marcrohloff: You're right,m I reverted this change

spec/grape_entity/entity_spec.rb
marcrohloff: I made the suggested changes and added several tests to verify that `ArgumentError` was raised when expected

lib/grape_entity/entity.rb
```diff
@@ -547,11 +547,14 @@ def ensure_block_arity!(block)
...
+      # Ensure that the function does not require any positional args
+      # (functions defined using `delegate` or `method_missing` take an arg of `*rest`
       arity = object.method(origin_method_name).arity
-      return if arity.zero?
+      required_positional_arg_count = arity >= 0 ? arity : -arity - 1
+      return if required_positional_arg_count.zero?
 
       raise ArgumentError, <<~MSG
-        Cannot use `&:#{origin_method_name}` because that method expects #{arity} argument#{'s' if arity != 1}.
+        Cannot use `&:#{origin_method_name}` because that method expects #{required_positional_arg_count} #{'argument'.pluralize(required_positional_arg_count)}.
```
numbata: The `.pluralize` is a ActiveSupport's method which requires `active_support/core_ext/string/inflections`.  It works today, but it depends on I18n config and inflection rules someone could override in a Rails app. How about `#{required_positional_arg_count == 1 ? 'argument' : 'arguments'}`? Zero magic, zero risk.

spec/grape_entity/entity_spec.rb
```diff
@@ -408,6 +408,22 @@ def method_with_one_arg(_object)
...
+
+            def method_with_required_and_optional_args(_required_arg, _optional1 = 1, _optional2 = 2)
+              'result'
+            end
+
+            def method_with_optional_args_as_a_splat(*_optional_args)
+              'result'
+            end
+
+            def method_with_required_args_and_an_optional_splat(_required_arg, *_optional_argds)
```
numbata:
```suggestion
            def method_with_required_args_and_an_optional_splat(_required_arg, *_optional_args)
```

lib/grape_entity/entity.rb
marcrohloff: I reverted to the original syntax instead. I also fixed the other typo

numbata:
Thanks for taking my comments into account, @marcrohloff! I added a few improvements on top of your commits before merging:

- Tightened the `Proc#to_s` regex to a strict anchored match to avoid false positives on _non-symbol-to-proc_ lambdas
- Fixed the arity math for variadic methods (`def foo(req, *rest)` was being miscounted)
- Added a bail-out for required keyword arguments so Ruby surfaces its native "missing keyword" error rather than a
misleading one from us
- Extracted `positional_arity_for` to keep the arity introspection logic self-contained and testable
- Expanded the spec coverage to include optional kwargs, splats, private methods, delegation-backed objects, and a canary spec for the `Proc#to_s` format the detection relies on

Great find, and thanks again for the PR! 👍

numbata (approved):

rc-marc-rohloff:
@numbata Thanks for cleaning that up

